PR #144, opened 2015-05-22
Merge changes: import/export breakpoints and line comments for the CPU/Disassembly view

Northern-Lights:
Let me know what you think of these enhancements.
1. Import/export breakpoints is implemented completely within the BreakpointManager plugin.
2. CPU/Disassembly view line commenting is implemented in QDisassemblyView and is accessed through Debugger.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ edb
 lib*.so
 .debug-shared/
 .release-shared/
-
+debugger.pro.user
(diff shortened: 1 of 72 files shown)


Northern-Lights:
I hate it when computers do things without my knowledge...  Anyway, good to know.  That solves my issue with the "Go to slot..." button in the Qt Creator UI designer.

PR #146, opened 2015-06-10
Merge enhancements

Northern-Lights:
Here is a brief list of enhancements in this pull request.  All are open for comment and discussion.

**CallStack.h/cpp**: A utility that examines the stack starting from rbp to determine the call stack by searching for return addresses.

**DebuggerCore.cpp**: Added a method to obtain the x86/64 names of the flag register.

**IDebugger.h/Debugger.cpp**: The flag representation string can now be right-clicked, and a flag bit can be selected for toggle.

**Debugger.cpp**: Rewrote the RunUntelRet class using a new algorithm to more effectively + accurately step out of functions.  (TODO: Step Out while sitting on a call initially steps into it).  Added Step Out button to UI toolbar (which simply calls the trigger action for Run Until Return).  Fixed issue with clicking "Cancel" in Comment entry QInputDialog (@eteran made changes to remedy this as well; my change covers the blank comment "" case).

**edb.cpp**: Simplified repaint() call for the cpuView.

**Backtrace plugin**: Displays the CallStack.  TODO: have the PC position be the top row, and the default "Run To Return" action only for the top row should be "Step Out".

**BreakpointManager plugin + QDisassemblyWidget**: BPM receives UI update signal from the CPU view so that the information in the plugin reflects breakpoints set/removed in the Debugger UI

**BreakpointManager plugin**: Fixed issue where internal breakpoints show up as blank lines.

**DialogBreakpoints.ui**: Made Export Breakpoints shortcut Alt + E.


eteran:
This is wonderful! There is quite a lot going on here, so bear with me while I sort through this and resolve  the few minor conflicts. All in all, looking very good though :+1:

DrOptix:
I did not look at all the improvements and diffs, but this one is a great improvement. Thank you!
> IDebugger.h/Debugger.cpp: The flag representation string can now be right-clicked, and a flag bit can be selected for toggle.

I'll do a pull when I get back home and I will test the new changes from master.

PR #147, opened 2015-07-18
Fix for FPU and MMX register presentation

10110111:
On x86 systems the original code made false assumptions about sizeof(long double) and layout of `fpregs_.st_space`. As a result, MMX registers showed garbage, and FPU ones were always zero. The first two commits fix this and thus allow to see nonzero (and correct) FPU registers as well as correct MMX registers.
Also, on both x86 and x86-64, `PlatformState::mmx_register()` assumed that `mm0-mm7` map to `st(0)-st(7)`, while in reality they map to `R0-R7`, which aren't organized as stack. The third commit fixes this mistake, thus showing correct results for instructions like `movd mm5,eax` independently of the value of `TOP` bits in FPU status word.


eteran:
Awesome. Thanks for the patch!

PR #150, opened 2015-07-19
Add support for reading SSE registers on x86 systems

10110111:
Following GDB, this commit makes use of `PTRACE_GETFPXREGS` instead of `PTRACE_GETFPREGS` on x86 targets. This way we get 1) support for reading SSE registers and 2) common format of `st_space` between x86 and x86-64.
Although in GDB there's also some code to check whether `PTRACE_GETFPXREGS` is supported, that code dates back to 1999 and is likely to be not needed these days. I've checked on Ubuntu 12.04, and `PTRACE_GETFPXREGS` works there too. So I assume any modern system with necessary version of Qt does also support `PTRACE_GETFPXREGS`.


eteran:
Another high quality patch! Thanks for your help

PR #160, opened 2015-07-20
Properly show memory expressions of sizes larger than ptrdiff_t

10110111:
This set of patches makes analysis view of instructions like `fld tbyte ptr [...]` and `movaps xmm0,xmmword ptr [...]` (and on x86 also `movq mm0, qword ptr [...]`) show correct values referred by the instructions.

I'm not sure though whether it's a good idea to use `uint32_t` for target on x86 and `uint64_t` on x86_64 — maybe use `uint64_t` on both, then the code would be identical (up to ordering of `switch` cases). But at least, the current way does work, so it's the matter of future changes.

### DIFF
--- a/src/arch/x86/ArchProcessor.cpp
+++ b/src/arch/x86/ArchProcessor.cpp
@@ -515,19 +515,34 @@ void analyze_operands(const State &state, const edb::Instruction &inst, QStringL
 				case edb::Operand::TYPE_EXPRESSION:
 					do {
 						const edb::address_t effective_address = get_effective_address(operand, state);
-						edb::address_t target;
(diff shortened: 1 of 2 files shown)


eteran:
I agree about using `uint64_t`. It would vastly simplify code in a few places if we just say "a pointer is up-to 64-bits" and used `uint64_t` everywhere a pointer could be used.

There has historically been a lot of code in both edb and edisassm which makes compile time decisions. So, we would just have to be careful to update any code which uses `sizeof(address_t)` or similar.

10110111:
Yeah, I think I'll make some arch-independent helper functions like `QString integerXX(void*)` to transform values of size `XX` to strings of their hex values.

Issue #159, opened 2015-07-20
FPU registers with exponent values exceeding those of float64 are shown as 0 or inf


10110111:
Here's an [example FASM code](https://pastebin.mozilla.org/8839973) and [checking code in C++](https://pastebin.mozilla.org/8839974). If running the fasm example, the first `fld` loads approximately `3.25435e-763`, which EDB shows as `0`. The second load is of approximately `-3.46264e+383`, which EDB shows as `-inf`.

Apparently, somewhere in the middle of processing the numbers are converted to double, thus losing representation range.

Running the C++ snippet you can check that the values are actually in the range of `float80`.

What's funny, GDB (at least 7.7.1) has exactly the same issue :)

10110111:
I'm taking this (dunno how to assign myself here).

eteran:
Awesome, just send a pull request mentioning this bug and I'll close when you're done :+1:

eteran:
I beleive that this was fixed by pull #163

PR #163, opened 2015-07-20
Fix precision and range loss when printing long double

10110111:
This fixes bug #159.
Please check that this builds on x86, I don't have access to such a system currently.

### DIFF
--- a/include/Util.h
+++ b/include/Util.h
@@ -19,6 +19,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef UTIL_20061126_H_
 #define UTIL_20061126_H_
 
+#include <QString>
+#include <sstream>
+#include <iomanip>
+#include <type_traits>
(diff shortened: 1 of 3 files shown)


eteran:
Thanks, I'll give it a test when I get home from work and merge if all goes smoothly. Just as a note, if you happen to have a multilib qt installed, it is possible to compile for either arch on the fly. For my system (gentoo). I can write:

    /usr/lib32/qt4/bin/qmake -spec linux-g++-32 # force a 32-bit build
    /usr/lib64/qt4/bin/qmake -spec linux-g++-64 # force a 64-bit build
    qmake # use system default of course

I don't know about other distributions, but it is probably possible there as well.

10110111:
Just checked and amended a bit on x86 system. Now it works.

eteran:
Patches look great. Keep it up the awesome contributions

PR #166, opened 2015-07-25
Remove a potential source of off-by-one or more serious errors.

10110111:
Sorry for that off by one error. I thought that approach would be more robust. This version must appear much better. Not sure how to deal with register view updater code though.


eteran:
No worries, it was easy enough to fix. I'm out of town til tomorrow night, so I'll take a look soon. Thanks!

PR #168, opened 2015-07-29
Add some functionality to `valueXX` types, use them in more places, including `reg_t` and `address_t`

10110111:
This set of patches allows to use generic functions like `toHexString()` or `signedFromString()` instead of hand-coding width, filling char etc. in much more places. Also it prepares the `valueXX` types to usage in a future X86 platform state, where it'll be useful to e.g. load value of XMM register from storage space of ZMM registers etc.
Also there are some minor cleanups like removing unnecessary conditional compilation, removing uses of deprecated `QString::sprintf()` functions, and a bit more.

### DIFF
--- a/include/Register.h
+++ b/include/Register.h
@@ -53,7 +53,7 @@ public:
 	QString name() const         { return name_; }
 	
 	template <class T>
-	T value() const              { return value_; }
+	T value() const              { return T(value_); }
 
 private:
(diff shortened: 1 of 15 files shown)


eteran:
these patches look great. Definitely inline with long term goals that I have. Keep up the great work!

Evan

PR #171, opened 2015-08-01
Add AVX support to PlatformState and a bunch of other improvements

10110111:
The main feature of this set of commits is that now AVX registers are displayed instead of SSE ones on systems where AVX is supported and enabled by the OS. Also the support for SSE-less systems was fixed. Now on such systems SSE registers are not displayed, but FPU ones still work correctly (this was broken by my `PTRACE_GETFPXREGS` patch before).

Another change is that invalid registers are no longer zeroed. Their values are instead marked by filling the corresponding memory with marker of `0xbad1`. This way 1) the user will immediately notice that the value being reported by EDB is untrustworthy and 2) the developer will see the bogus values when doing test runs. This also means that the functions can't rely on invalid `Register` having zero value. They instead should explicitly check validity of `Register` they received.

The above change also makes the previous situation that the registers were showed although zero even when there were no program attached to. Now in these cases (more precisely, when State is empty) the registers are not displayed at all.

Also, now general purpose registers are displayed in a bit more "natural" order, namely according to their architectural index (like index `i` in `0xB8+i` instruction).

Also I've replaced previous hard to use `string_hash`, which required me to separate strings into chars, by a new one, which accepts string literals (up to 8 chars long too) and has an interface of a function, rather than `struct`.


eteran:
love the usage of `constexpr` to improve the `string_hash` function. The migration to c++11 was a relatively recent switch. Loving the code improvements that result :-)

PR #181, opened 2015-08-03
Fix hex number presentation in x86_64 FPU view too

10110111:
I forgot to do this for x86_64 ArchProcessor in the previous pull request.


eteran:
All the more reason to work out the best way to share code between related arches! Thanks!

PR #210, opened 2015-08-29
Replace edisassm with Capstone

10110111:
This set of commits implements a wrapper around Capstone, enabling several features like AT&T syntax and dynamic switching of disassembler bitness.

I've decided to not wait until I fully implement non-native bitness support, because upstream changes have begun to conflict with this branch, and resolving the conflicts is not much fun.

The non-native bitness support is a separate feature anyway, and involves a bunch of changes, so it should in fact be kept separate from this pull request.

NOTE: I've not removed any of edisassm sources from the tree, and `Formatter.{cpp,h}` in `capstone-edb` directory are just placeholders, but I think this cleanup would better be in the next pull requests (and removing edisassm from the tree would be better if done by you since I'm not sure how git works with submodules).


eteran:
Doesn't `src/src.pro` need a `LIBS += -lcapstone` in there?

10110111:
Oh, indeed. I forgot that I was compiling it having once done `qmake PREFIX=/usr LIBS+=-lcapstone`. I'll fix now.

10110111:
Done.

eteran:
OK, I think that this pull request is worthy of a merge! I plan to do a little cleanup afterwards, so please be sure to keep an eye out of any small updates that I do afterwards :-)

PR #300, opened 2015-11-17
Let the user enter program arguments and working dir in file open dialog

10110111:
This clones the functionality present in OllyDbg's file open dialog. OllyDbg seems to do some ugly-looking hack on the stock Win32 dialog, so that the entries are misaligned with the stock ones. This attempt tries to use QFileDialog's layout to align the labels and entries so that the result looks much better.

A downside of this feature is that we no longer have native dialogs since Qt4+ doesn't allow us to add custom widgets using the official API.

Another downside is that no one guarantees that the layout will always be as it is. OTOH, I see no reasons why they would suddenly change it, and the code does sanity check the dialog's layout and refrains from modifying if it looks suspiciously.

### DIFF
--- a/src/Debugger.cpp
+++ b/src/Debugger.cpp
@@ -28,6 +28,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "DialogOptions.h"
 #include "DialogPlugins.h"
 #include "DialogThreads.h"
+#include "DialogOpenProgram.h"
 #include "Expression.h"
 #include "IAnalyzer.h"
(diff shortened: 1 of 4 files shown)


eteran:
Lemme take a look at this code and try it out to see how it works (I suspect well :-)).

10110111:
Any news on this PR?

eteran:
Sorry, works been busy, I'll give it a try some time today and probably merge it shortly there after :-)

eteran:
this looks great.

PR #315, opened 2015-11-26
Use QStyledItemDelegate as base for RegisterViewDelegate

10110111:
Currently `RegisterViewDelegate` is based on `QItemDeletage`, which makes register view look not fully styled. Namely, with Oxygen, we lose hover effect and all those non-flat selection rectangles.
Default `QTreeView` and similar views install `QStyledItemDelegate` as their renderers. So `RegisterViewDelegate` should actually be based on it to make untouched rows look correct.

BTW, do you think it'd be a good idea to use `QStyle::drawPrimitive()` for disassembly, stack and hex views, so that they allowed hover effects and styled selections?

### DIFF
--- a/src/widgets/RegisterViewDelegate.cpp
+++ b/src/widgets/RegisterViewDelegate.cpp
@@ -23,7 +23,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 // Name: RegisterViewDelegate
 // Desc:
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


eteran:
This is a good patch. Funny thing, I based that code on code that the Qt folks themselves wrote! It is based on the code for the qt-designer widget picker.

10110111:
Right, `QStyledItemDelegate` appeared in Qt 4.4, so `QItemDelegate` was the only option before that.

PR #318, opened 2015-12-08
Fix UB in string_hash<9> due to left shift by 64

10110111:
The default argument `n=N` makes the first shift `..<<(8*(n-1))` result in UB when `N==9` (i.e. string with maximum length of 8), since the RHS evaluates to 64, equal to number of bits in `uint64_t`. This commit fixes it to start from `n=N-1`: the last character is the NULL terminator anyway, so no point in processing it.

### DIFF
--- a/include/string_hash.h
+++ b/include/string_hash.h
@@ -26,11 +26,11 @@ namespace edb {
 
 namespace detail {
 
-template<std::size_t N, std::size_t n=N>
+template<std::size_t N, std::size_t n>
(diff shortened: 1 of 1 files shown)


eteran:
Good catch, maybe we should use a static assert too?

10110111:
I'm not sure how you can have static assertions in C++11 constexpr functions.

eteran:
I was originally thinking of a static_assert on the template parameter... but looking at it, it is completely unnecessary since we use enable_if.

PR #328, opened 2016-01-12
Lower required CMake version

10110111:
These commits let one use CMake 2.8 to compile EDB. At least this works on my system with CMake 2.8.12.2. May need some additional tweaking for 2.8.7 (which is current on Ubuntu 12.04 LTS). Additional commit also makes use of pkg-config to find Capstone, so that the users who have it installed in non-default location will still have EDB compile, provided they have set `PKG_CONFIG_PATH` correctly.

Please check that this doesn't break with CMake 3, since I don't have it installed anywhere to test.

BTW, is it only me, or does cmake install plugins not in `${CMAKE_INSTALL_PREFIX}/lib64/edb` on 64-bit systems? For me it installed them to `${CMAKE_INSTALL_PREFIX}/lib/x86_64-linux-gnu/edb` instead on Kubuntu 14.04 amd64. In any case, default plugin search dirs may need some revision.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 2.8)
 
 # if the user has not specified which version, we try to make Qt5 happen
 # if they have, we prefer their preference
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ project (edb)
 include("GNUInstallDirs")
 
 find_package(Boost 1.35 REQUIRED)
+find_package(PkgConfig REQUIRED)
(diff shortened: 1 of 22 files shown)


10110111:
Oops, wrong branch.

eteran:
So regarding the installation location of plugins. It is all based on the GnuInstallDirs module. 

If for some reason it happens to give the wrong directory, I think I'd have to consider that a bug in the system configuration. Since that's where the system is telling us to install things, it's probably where we should prefer unless overridden by the user since the distribution may have their own reasons for using unconventional directories.

10110111:
Yes, I found that if I delete `edb.conf`, EDB finds plugins normally, that's why I've removed this part of text in updated pull request.

PR #330, opened 2016-01-12
Lower required CMake version

10110111:
These commits let one use CMake 2.8 to compile EDB. At least this works on my system with CMake 2.8.12.2. May need some additional tweaking for 2.8.7 (which is current on Ubuntu 12.04 LTS). Additional commit also makes use of pkg-config to find Capstone, so that the users who have it installed in non-default location will still have EDB compile, provided they have set `PKG_CONFIG_PATH` correctly.
And one more commit makes proc-pid-mem-writes-test avoid regenerating the file, so that cmake didn't have to rebuild dependee on every `make`.

Please check that this doesn't break with CMake 3, since I don't have it installed anywhere to test.


10110111:
OK, so now it is compatible with CMake down to 2.8.7.

PR #333, opened 2016-01-14
Fix adding of resources with CMake 2.8

10110111:
Don't know whether it does work with CMake 3, but with 2.8 I found that the current instruction arrow in disassembly view was missing, and list of assemblers in plugin settings was empty and disabled. This commit fixes adding Qt resources with CMake 2.8.


10110111:
OK, now fixed to work with Qt5.

PR #367, opened 2016-04-21
Check that index register is a GPR, otherwise EDB'd crash. 

10110111:
Currently EDB crashes when it encounters an instruction with VSIB addressing, e.g.

    C4 E2 65 93 8C 79 99 33 55 00   vgatherqps xmm1,dword [ecx+ymm7*2+0x553399],xmm3

First commit works around this, leaving addition of VSIB support as a TODO.
Second commit prevents senseless calculations when e.g. VSIB has `ymmN` as index while the CPU/OS doesn't support AVX.


eteran:
good catch

PR #369, opened 2016-04-23
SIMD floating-point operands formatting

10110111:
This enables quick analysis view display floating-point operands of SIMD instructions in more useful format. E.g. if you have an instruction like

    vrcpss xmm3,xmm7, dword [esp]

you'll have the following lines in quick analysis view:

    xmm7={124.352, 972e+3, 0.0, -5.23526}
    xmm3={0.0,0.0,-0.0,1.3}
    dword [esp] = [0xffe3ce50] = 1.40129846e-45

according to the types of the operands (here xmm7 and xmm3 are packed single-precision, while `[esp]` is scalar single-precision floating-point numbers).

Also FPU BCD operands are now formatted according to BCD format too.

### DIFF
--- a/include/Register.h
+++ b/include/Register.h
@@ -62,6 +62,7 @@ public:
 	Type type() const            { return type_; }
 	QString name() const         { return name_; }
 	std::size_t bitSize() const  { return bitSize_; }
+	const char* rawData() const { return reinterpret_cast<const char*>(&value_); }
 	
 	template <class T>
 	T value() const              { return T(value_); }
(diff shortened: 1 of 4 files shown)


eteran:
Looks like this uses stuff in capstone which isn't in 3.0.4. Can we verify which version will be required for this? Or perhaps wrap some of the code in an `#ifdef` which tests capstones version?

10110111:
Oh, didn't notice. I was using capstone's `next` branch, since `master` has quite some bugs which have long been fixed in `next`. I'll try to see what's missing in `master`.

10110111:
OK, now this is fixed.

PR #380, opened 2016-04-27
Avoid processing all events just to repaint the status bar.

10110111:
`processEvents()` appears to take quite a bit of performance away from EDB (especially visible on my EEE PC with Intel Atom).
This commit uses the least necessary action to make sure the status is repainted, so as to avoid too much time spent uselessly.


10110111:
Oh well, this appears to not work well for some reason. Retracting for now.

10110111:
Seems fixed with new commit.

PR #386, opened 2016-04-29
New options for instruction formatting

10110111:
This adds some options for instruction formatting

1. Alignment of operands (tab-like; two tab stops for short and long instructions)
2. Avoiding extraneous display of local module names in jump/call targets (e.g. `<_dl_start>` instead of `<ld-2.14.1.so!_dl_start>` when called from within `ld-2.14.1.so`)
3. Transformation of `[rip+0xd1521ace]` into `[rel 0xad7e55]`.


10110111:
Travis seems to be quite volatile. Sometimes it does succeed in installing the required packages, other times it fails...

PR #412, opened 2016-05-10
RFC: Only syntax-highlight disassembly if the instruction isn't selected

10110111:
Otherwise we have bad readability due to low contrast. I don't have better ideas than to disable syntax highlighting for selected line. See the comparison:


eteran:
I think this is a real improvement. I like it and honestly it's similar to what many text editors do.

10110111:
I think the same thing should be done with address. But the address should first get some support for real theming instead of hard-coded red color — and this, in its turn, should somehow be done consistently with QHexView.

PR #418, opened 2016-05-16
Add some helpful debugging options

10110111:
This adds some options which are often helpful when debugging often-restarted programs: to disable ASLR and to disable lazy binding of dynamically linked functions.


eteran:
:+1:

PR #440, opened 2016-08-13
Fixed missing std:: qualifier, and cmath includes in graph source.

MrDetonia:
`<cmath>` includes were missing from `GraphNode.cpp` and `GraphWidget.cpp` in `src/graph`
also fixed a missing `std::` in `GraphNode.cpp`
Both of these issues were breaking the build.

### DIFF
--- a/src/graph/GraphNode.cpp
+++ b/src/graph/GraphNode.cpp
@@ -26,6 +26,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <QPainter>
 #include <QPainterPath>
 #include <QAbstractTextDocumentLayout>
+#include <cmath>
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


10110111:
Interesting... has GCC stopped pushing cmath functions into global namespace?

MrDetonia:
GCC doesn't, but G++ should. It confused me as well, but this seems to fix it.

I'd like to hear if anyone else was having this issue as well when building, perhaps the `#include <cmath>`s can be removed from this pull request.
The missing `std::` was definitely causing an error though.

10110111:
I've had some similar issues in other projects with newer GCC versions (by GCC I mean GNU Compiler Collection, not GNU C Compiler). Upstream seems to try to make C++ headers cleaner, so that each standard header included fewer others, that's why these compilation failures occur.
So I think this PR does have a point.

PR #445, opened 2016-09-28
install logo file as edb.png

ZeroChaos-:
The desktop file expects an icon called edb, not edb48-logo

https://github.com/eteran/edb-debugger/blob/master/edb.desktop#L6


ZeroChaos-:
ugh, this doesn't work, the file has to actually be renamed

PR #448, opened 2016-09-29
Add an Instruction Inspector plugin.

10110111:
The new plugin displays Capstone-supplied info about the selected instruction.

It also allows to compare output of multiple disassemblers with Capstone's output. The disassemblers for comparison are (when installed in the system): `ndisasm`, `objdump` and Agner Fog's `objconv`.


eteran:
Hmm, it looks like `x86_xop_cc` isn't in the version of capstone that our test scripts use.

10110111:
Ah, right... Will look for a fix.

eteran:
Yea, looks like 4.0-alpha1 was the first version to that that `enum` Perhaps we can just check `CS_API_MAJOR ` and `CS_API_MINOR ` and conditionally support printing instructions of that type?

10110111:
Yeah, it seems we already do this in `Instruction`.

10110111:
Should now work.

eteran:
I plan to merge once 0.9.21 is tagged and we get the grsecurity/gentoo ebuild stuff sorted. Thanks!

PR #458, opened 2016-10-10
Build and upload AppImage

probonopd:
Build and upload an [AppImage](http://appimage.org/) for Qt 5.6 builds (builds prior to  Qt 5.5 are not yet bundled correctly due to what I suspect to be a [linuxdeployqt](https://github.com/probonopd/linuxdeployqt) bug). The URL of the built AppImage can be seen in the build log.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ env:
     global:
         - MAKEFLAGS="-j2"
     matrix:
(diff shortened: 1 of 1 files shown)


eteran:
Interesting, I had never heard of AppImage. Sure, we can make an automatic bundle of each commit :-)

PR #459, opened 2016-10-12
Add completer for Go-To dialog

knikolenko:


eteran:
Awesome, thanks!

10110111:
Won't this be awfully slow for large programs with lots of symbols? E.g. Symbols dialog already is quite slow in such cases.

eteran:
Possibly. I can say that edb debugging itself is not too slow. edb isn't huge, but it isn't tiny either.

If it does turn out to be slow in some cases, I would consider that an issue worth filing in the tracker and exploring ways to make things faster.

knikolenko:
Can create a QList of symbols in real time in the SymbolManager, instead of creating a list each time you call the GoTo dialog?

Issue #237, opened 2015-09-16
No easy way to go to previous/next location


10110111:
I do understand there're bookmarks, but they are quite unhelpful when you just want to jump to previous/next location. If EDB maintained a stack of recent locations or something like that, it'd be much more useful.
Something like this is implemented in OllyDbg, it has shortcut keys `Num+`/`Num-` for next/previous location.

eteran:
Is there a particular definition for "location"? do we mean previously selected CPU addresses? previously stopped on addresses? Is the list of previous addresses cleared when the debugee is resumed?

10110111:
If mimicking OllyDbg functionality, here's what it says in its help on Command history:

> OllyDbg keeps track of up to 1000 last commands displayed in Disassembler window. Each time the debugged program stops on breakpoint or exception, or you step next command, or follow address of jump or call, or simply press Return, OllyDbg adds current address together with thread identifier to the command history. If run trace is inactive, you can navigate through the history by pressing buttons '+' and '-'.

eteran:
A recent merge adds navigation of previous locations via plus/minus hotkeys and also updates the selection to the current location on each stop.

PR #461, opened 2016-10-13
Add navigation history for CPU view

knikolenko:
Saves the last 100(customizable) movements.
Keys for navigation: Enter, Plus, Minus.
Fix scroll up/down to navigate the lines


eteran:
Very nice, i'll look at this a bit later tonight. Do you feel that this addresses issue #237 ?

PR #467, opened 2016-11-16
Correctly handle whitespace in file paths in map files

pgkos:
This is a correction of a small bug which caused the file paths with spaces (parsed from map files) to be truncated (before the first space).


eteran:
This looks pretty good. Thanks!

PR #469, opened 2016-11-18
Always use /proc/<pid>/mem over ptrace whenever possible

AaronOpfer:
This fixes the infuriating issue where it was impossible to add breakpoints, remove breakpoints or edit the dump while a debuggee was running. This was caused by both read_bytes and write_bytes putting single byte read/writes (one of each is necessary to make a breakpoint) on a so-called "fast path" which tries to avoid /proc/<pid>/mem.

While fixing this issue I decided this code was a little too "patch-fixy" for me. It looks like we patched in /proc reading to fix limitations occasionally but were not consistent in where we applied these patches. As a result we sometimes we check /proc, sometimes we check ptrace, and sometimes we do both. I decided to refactor the area so that /proc always comes first followed by ptrace, and I renamed each method to more clearly indicate its responsibility. Since all this refactoring only touches the private API I think this should be pretty safe.

### DIFF
--- a/plugins/DebuggerCore/unix/linux/PlatformProcess.cpp
+++ b/plugins/DebuggerCore/unix/linux/PlatformProcess.cpp
@@ -192,8 +192,20 @@ QList<Module> loaded_modules_(const IProcess* process, const std::unique_ptr<IBi
 // Name: PlatformProcess
 // Desc: 
 //------------------------------------------------------------------------------
-PlatformProcess::PlatformProcess(DebuggerCore *core, edb::pid_t pid) : core_(core), pid_(pid) {
-
+PlatformProcess::PlatformProcess(DebuggerCore *core, edb::pid_t pid) : core_(core), pid_(pid), ro_mem_file_(0), rw_mem_file_(0) {
+	if (!core_->proc_mem_read_broken_) {
+		QFile* memory_file = new QFile(QString("/proc/%1/mem").arg(pid_));
+		auto flags = QIODevice::ReadOnly | QIODevice::Unbuffered;
+		if (!core_->proc_mem_write_broken_) {
+			flags |= QIODevice::WriteOnly;
+		}
+		if (memory_file->open(flags)) {
+			ro_mem_file_ = memory_file;
+			if (!core_->proc_mem_write_broken_) {
+				rw_mem_file_ = memory_file;
+			}
+		}
+	}
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


eteran:
I'll take a look at the patches ASAP. 

The concern was that opening the virtual file in proc, seeking, doing I/O, etc.. was going to be slow for single byte operations. So we'll have to look at how this performs.

Thanks!

eteran:
BTW, this merge seems to include the stuff for pull request #468 . Any chance we can separate them out?

AaronOpfer:
I'm hoping that two syscalls (seek + read) is better on average than one syscall per word. Probably not for single bytes, but I hope it's faster when it counts the most (dumping large amounts of data). It's a shame that mmap isn't supported on the remote process' memory so that we could avoid kernel->user copies.

eteran:
@10110111 So the original code too measures to try to deal with address values > 32-bits on a 32-bit debuggee. I *think* this code still deals with that OK since it will always prefer `/proc/<pid>/mem` access and the `ptrace` approach either works or it doesn't. But I ran into this issue a lot less than you did.

What do you think?

10110111:
I'm not sure that it handles this case gracefully. Namely, if `EDB_IS_32_BIT && address>0xffffffffULL`, this patch will truncate the address and read/write garbage via `ptrace`, since the check has been deleted completely. I think after these changes the check should still be done, and if it results in `true`, then just return failure, since `/proc/.../mem` must have failed if we got to use of `ptrace`.

AaronOpfer:
Maybe the user should be given a warning if read /proc/<pid>/mem is broken and they're debugging a 64 bit process, or maybe even immediately detach. I fail to see how the debugger can accomplish much in that situation. We can still add the check around the write ptrace, though.

eteran:
@AaronOpfer Yea, I think we should keep the check in there.

10110111:
> Maybe the user should be given a warning if read /proc//mem is broken and they're debugging a 64 bit process, or maybe even immediately detach. I fail to see how the debugger can accomplish much in that situation.

You don't take into account x32 processes. They still have 32-bit pointers, but 64-bit address space, so e.g. assembly code could do some 64-bit calls and access higher memory, although by default 32-bit EDB would work normally with them even without `/proc/.../mem`.

AaronOpfer:
Okay, I've added the checks back in. Let me know if you want me to squash these commits before merges.

eteran:
This looks good, in its current form. I'll do some tests and probably merge it later.

Thanks!

PR #471, opened 2016-11-20
Annotate constants in disassembly if they appear to point to ASCII

AaronOpfer:
This patch-set annotates constants in the disassembly screen which appear to point to ASCII strings.

image

An unfortunate limitation of my patchset is that it does not understand the constants inside relative expressions. This appears common on x64 targets. Here's a simple hello world showing that the two LEAs that are RIP-relative are not annotated:

image

If anyone has any ideas how I can fix this corner case, I'm all ears.

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -994,11 +994,12 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 					ascii_address = oper.relative_target();
 				} else if (
 					oper.type() == edb::Operand::TYPE_EXPRESSION &&
+					oper.expression().index == edb::Operand::Register::X86_REG_INVALID &&
 					oper.expression().displacement_type == edb::Operand::DISP_PRESENT)
 				{
 					if (oper.expression().base == edb::Operand::Register::X86_REG_RIP) {
 						ascii_address += address + oper.owner()->size() + oper.expression().displacement;
-					} else if (oper.expression().displacement > 0) {
+					} else if (oper.expression().base == edb::Operand::Register::X86_REG_INVALID && oper.expression().displacement > 0) {
 						ascii_address = oper.expression().displacement;
 					}
 				}
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -870,13 +870,10 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 			}
 		} else
 
-#if 1
 		// highlight header of binary
 		if(binary_info && address >= region_->start() && address - region_->start() < binary_info->header_size()) {
 			painter.fillRect(0, y, width(), line_height, QBrush(Qt::lightGray));
-		} else
-#endif
-		if(row_index & 1) {
+		} else if(row_index & 1) {
 			// draw alternating line backgrounds
 			painter.fillRect(0, y, width(), line_height, alternated_base_color);
 		}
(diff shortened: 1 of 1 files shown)


10110111:
+				if (address > 0x10000) { // FIXME use page size

Why do you really want to ignore small addresses? One can `mmap` to those addresses if `$(sysctl vm.mmap_min_addr)` allows this.

> An unfortunate limitation of my patchset is that it does not understand the constants inside relative expressions. <...> If anyone has any ideas how I can fix this corner case, I'm all ears.

You want to look at `get_effective_address()` in `ArchProcessor.cpp`. Namely, grep for `X86_REG_RIP`.

AaronOpfer:
Thanks for the pointer to `ArchProcess.cpp`. Now those constants are decoded as well.

> Why do you really want to ignore small addresses?

My worry was that very common integers like 1, 2, 255, 256, 512, etc would appear a very large number of times in disassembly output in some situations (array/struct initialization or CRC routines come to mind). If I looked up ASCII strings for all of them it would just waste rendering speed on nothing in 99.9% of configurations.

src/widgets/QDisassemblyView.cpp
```diff
@@ -989,18 +989,25 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
-					// TODO Interact with capstone to evaluate operand's displacement
+					ascii_address = oper.relative_target();
+				} else if (
+					oper.type() == edb::Operand::TYPE_EXPRESSION &&
+					oper.expression().displacement_type == edb::Operand::DISP_PRESENT)
+				{
+					if (oper.expression().base == edb::Operand::Register::X86_REG_RIP) {
+						ascii_address += address + oper.owner()->size() + oper.expression().displacement;
+					} else if (oper.expression().displacement > 0) {
+						ascii_address = oper.expression().displacement;
```
10110111: Won't this give wrong results for code like `mov eax, [ecx+edx*4+disp]` (interpreting it as if it was `mov eax,[disp]`)?

10110111:
Argh, github's UI is unintuitive...

src/widgets/QDisassemblyView.cpp
AaronOpfer: Depends on your definition of wrong. :smile: We could substitute the current values in those registers and solve the address,  but unless you are paused on that instruction those results are probably "wrong" as the relevant registers are probably going to be garbage values until the routine fixes them right before the instruction runs. Honestly I was only really worried about the RIP relative and the constant ASCII as they were low-hanging fruit and hadn't thought through this much.

10110111: That's why I suggested to add checks that index and base are empty. If we look at OllyDbg, BTW, it just adds red "(current registers)" in such situations, but this thing in the comment column only appears if you select the line.

AaronOpfer: Okay; I've updated the PR so that it requires empty index, and empty base in the non-RIP case.

AaronOpfer:
I think it probably makes a lot of sense to split the "If there's a string at this address let's make a new string called 'ASCII <%s>'" logic into its own method, especially because various plugins/core parts appear to have this exact same logic in it. I see the behavior in the ODdbgRegister plugin. I also see in the HeapAnalyzer plugin, but it also does signature detecting for PNG, XPM, BZIP, GZIP and COMPRESS (for some reason). We should be implementing this "Address to String representation of address contents" logic in a single place, especially because we probably want to expand this behavior to detect UTF-16 strings since they're heavily used in Windows.

eteran:
@AaronOpfer, @10110111 What's you guy's thoughts on the maturity of this patch?

BTW: edb has a function to find UTF-16 strings as well: `edb::v1::get_utf16_string_at_address`, so it's pretty easy to add that in.

I do agree that we probably want to centralize a common implementation of "binary blob analysis". Maybe there should be some module that is extendable via a plugin, where you give it a block of bytes, and it gives you back an object representing what it "thinks" that thing is? I'll have to think about how that would work...

AaronOpfer:
I'm confident the code works as it is now. I think centralizing binary blob analysis is a good idea, but I don't have much appetite for doing it myself. Maybe when I'm a little warmer with the codebase.

eteran:
OK, well i love the feature, so I'll merge it in 👍

PR #472, opened 2016-11-21
Fix incorrect entry point detection for ELF32/64

AaronOpfer:
ELF32 and ELF64 incorrectly return the offset to the entrypoint instead of the actual entrypoint address (in my example it tried to add a breakpoint to 0x580 instead of 0x55550580). This breaks `Debugger::set_initial_breakpoint`'s attempt at setting a breakpoint on the application's entrypoint, forcing the developer to step through the linux loader code for a long time.

### DIFF
--- a/plugins/BinaryInfo/ELF32.cpp
+++ b/plugins/BinaryInfo/ELF32.cpp
@@ -78,7 +78,7 @@ bool ELF32::native() const {
 edb::address_t ELF32::entry_point() {
 	read_header();
 	if(header_) {
-		return header_->e_entry;
+		return header_->e_entry + region_->start();
 	}
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


AaronOpfer:
Okay, it looks like I don't have the whole story. My executables compiled by gcc6 require this offset, my executables compiled by an older gcc5 do not. It looks like RIP-relative offsets allow new kinds of position-independent ELF files.

eteran:
So it's a little unclear from the comment thread, is this patch considered to be in good shape for merge? Or does it need more work?

AaronOpfer:
It's not ready yet. We need to figure out what is the base address a module expects to be loaded at, compare it to the actual address it was loaded at, and then subtract them to come up with an correction we apply to the entrypoint value from the ELF header (which is a virtual address).

eteran:
OK, I'll close this pull then. Feel free to push it again when it is more ready for the prime time 😀

PR #474, opened 2016-11-23
Make ELF32/ELF64 use Templates

AaronOpfer:
Because I foresee doing a lot of work in this area to fix issues I've raised, I decided to refactor ELF32 / ELF64 into one base template with specializations in order to reduce the dramatic amount of copy and paste that was going on.

This branch contains no other attempts at bugfixes, although there are some whitespace fixups.

This mainly works thanks to adding extra typedefs/enums to `elf_header.h`, which was copied from GNU... hopefully modifying our copy isn't too big of a deal. If it is, we would have to extend their structs instead of rewrite them which could cause a lot of ugly casts in various places to keep API compatibility (maybe; I didn't try).


eteran:
I generally like the idea of using a template for this. Gonna take look at patch before merging though because there is a lot going on there :-).

eteran:
played around with it, looks good!

PR #475, opened 2016-11-23
Fix /proc/<pid>/stat parse failure when binary name has .

AaronOpfer:
This fixes an issue where my executable named "test2.5" would fail to have any entrypoint detected for it at all. Turns out this was because the parsing of /proc/<pid>/stat was failing its match expression which has a whitelist of sane characters to expect.

I don't feel like this whitelist is a good idea because  binaries can have any name in unix as long as it doesn't contain a NUL or / character. But I'll do the cheap and easy fix for my case instead.


eteran:
Yea, the white list is tough because this field can have basically any side and can contain any character. If we can develop a better solution, I am definitely interested. In the mean time, this patch will do :-)

PR #477, opened 2016-11-23
Fix ELF entrypoint detection failing on relocated binaries

AaronOpfer:
This is the new version of #472.

~~Please inspect the "Refactor IBinary & Fix ELF relocation bug" commit directly as this changeset is based on my other pull request #474 which is not the code under inspection here.~~

First off: I've broken plugin API. I did this because I thought creating an object and then calling another method to determine whether the object should actually work at all (`validate_header`) was a bit backwards. If the object is invalid, and will never become valid, it should never have existed in the first place. So:

 * The entire contents of `read_header` have been moved into the constructor.
 * `ELFXX`'s constructor will now raise an exception when the read/write fails, ELF header is invalid, etc..
 *  `validate_header` no longer makes sense as public API because we call it from the constructor. It is now private.
 * All of the member functions no longer invoke `read_header` on every call or test `header_` for null. In fact, `header_` is not even a pointer at all anymore.
 * `header_size()` will now add the size of of the program headers to the value it returns if the PHs are directly following the ELF header. This just makes the gray box denoting the header in the disassembler view a little bit longer. Just an aesthetic thing for me.
 * I also changed edb's `q_BinaryInfoList` to be reordered every time a binary loader other than the first one works successfully so that the successful loader will be attempted first next time. I did this when I noticed the ELF32 classes were being instantiated and thrown away and then succeeding with ELF64 immediately afterward on my 64bit debuggee. Now, as soon as the 64 bit one works, it will be used first every time after. This should make the binaryinfo objects slightly cheaper to retrieve (but we should be caching them).

And finally the bug fix: in the `ELFXX` constructor, we iterate the program headers and look for any LOAD header. We then take the LOAD header with the lowest address and assume it indicates the base address as suggested on [this SO post](http://stackoverflow.com/questions/18296276/base-address-of-elf) (we have a few sanity checks in here so that strange binaries will hopefully make us fallback to the old behavior). Then, finally, `entry_point` now adds the memory region start address and subtracts the in-memory base address. 

The part of this code review I am the last confident about is the C++ exception classes and handling as I am mostly unfamiliar with them, so I would appreciate feedback on those areas especially.


eteran:
Thanks,

I especially like the optimization you made in your last point.  Once #474 is merged in, this one will follow shortly. As a note, I am traveling for the holiday, so I may not get to it until I get back on Sunday.

Thanks again!

AaronOpfer:
I've rebased my changes on master so now this pull request should be just the refactor / ELF entrypoint fix.

PR #480, opened 2016-11-29
Add CPU Region information above Disassembly view

AaronOpfer:
This adds a little "you are here" indicator above the CPU region view. The idea is that this will help you tell where you are.

image

I'm pretty happy with this except that I can't figure out why my label and the "Registers" label to the right are not vertically aligned with one another. That's a bit nit-picky though so I'm not going to hold up my PR over it.

### DIFF
--- a/src/Debugger.cpp
+++ b/src/Debugger.cpp
@@ -2267,7 +2267,6 @@ void Debugger::update_tab_caption(const std::shared_ptr<QHexView> &view, edb::ad
 // Desc:
 //------------------------------------------------------------------------------
 void Debugger::update_data(const DataViewInfo::pointer &v) {
-
 	Q_ASSERT(v);
 
 	const std::shared_ptr<QHexView> &view = v->view;
@@ -2356,7 +2355,6 @@ IRegion::pointer Debugger::update_cpu_view(const State &state) {
 // Desc:
 //------------------------------------------------------------------------------
 void Debugger::update_data_views() {
-
 	// update all data views with the current region data
 	for(const DataViewInfo::pointer &info: data_regions_) {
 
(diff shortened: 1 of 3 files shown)


10110111:
This takes away screen space, which is especially important on small wide monitors like on netbooks. IMO it should at least be possible to turn it off. The EIP/RIP comment in the Register View already says where you are.

AaronOpfer:
I am mentally looking for the module name I'm working with at the top of the CPU view or the window title bar because that's what ollydbg did; in olly if the CPU view is maximized the current module is in the window title bar, else it's in the CPU window title bar. That's why I wanted to put the label here. Would you prefer we move the current module name into the title bar and remove this label?

10110111:
Title bar currently shows path to the executable and PID. I'm not sure it's a good idea to replace this with current location in the debuggee.

AaronOpfer:
Actually, the value I'm displaying is the current view, not the RIP. So if
you navigate to another module the value I display changes whereas the
register view does not.

On Tue, Nov 29, 2016 at 1:06 PM, Ruslan Kabatsayev <notifications@github.com

eteran:
So, I imagine that the label doesn't line up because the "Registers" is actually the title bar of a dock window, which I'm sure has a few margin adjustments done by Qt by default to prevent it from being too spaced out. It should be in principle possible to get perfect alignment, but I agree that it's not a showstopper.

I do agree with @10110111 that since it changes the amount screen realestate used, that we should consider this carefully and think about whether or not there is a better place to show this info. That being said, I do think that it is useful information. So we'll have to develop a good compromise between having the info and having more space.

As mentioned, perhaps having it be hidable is good enough.

eteran:
Minor nitpick on the PR, it's trivial enough to fix going forward. But please wrap all string constants which are user UI facing with `tr`, this will allow internationalization in the future.

As a bonus, it results in a `QString`, so things like this:

`QString("CPU: [%1-%2] %3").args(...)`

can be:

`tr("CPU: [%1-%2] %3").args(...)`

AaronOpfer:
Does qt automatically change margins and padding on different DEs? I would hate for my adjustments that look right on my screen to look wrong on someone else's. In any case I reduced the padding significantly (first time messing with qt designer).

image

I also have experimented with making the label use the CPU font, but it looks kinda ugly to me:

image

10110111:
> Does qt automatically change margins and padding on different DEs?
It's not Qt in fact, it's the widget style who gives margins to Qt. This most likely does differ from one style to another. You can of course see if Qt does have adjustable margins in the exact widget you're interested in — maybe dock titlebars aren't affected, I'm not sure.

AaronOpfer:
What do we think about having the currently viewed module name in the window title instead? We could truncate the executable path to just the filepart to make some room for the current module name (if any).

AaronOpfer:
I'm just going to close this because I don't think anyone else wants it, at least not like this. Maybe if/when inspiration strikes I'll have a better approach for this.

eteran:
Fair enough, but please don't be discouraged. You contributed a lot lately and it really has been a huge help to the project!

PR #481, opened 2016-11-29
Change Analysis Bar message

AaronOpfer:
* No longer display "No Analysis for This Region" if region is not valid (such as before any debugee has been launched)
 * Change "No Analysis For This Region" to instead say "[libname]: No Analysis Found" which might make it easier to understand why Analysis disappeared while stepping through code that invokes another library.


eteran:
Looks good!

PR #483, opened 2016-12-01
When reljmp selected, draw line to jmp destination

### LINKED ISSUE #464: Add jump destination line-drawing to disassembly view
It's probably easiest to explain with this screenshot of ollydbg, my old time favorite.Notice the red line visible on this page:

http://www.ollydbg.de/Pics/OllyDbg2.gif

When the jump instruction is highlighted, the destination of the jump is drawn in the disassembly view with a connecting line. Without this feature, it is very difficult to find my way around a lot of conditional code in edb.

AaronOpfer:
Fixes #464.

This draws a line when highlighting a jump toward the jump destination.

image

It's not completely analogous to ollydbg because olly will also highlight jumps that point TO the current instruction. Adding that feature will require a big refactor of the QDisassemblyView to dissasemble a large number of instructions ahead of time and then perform analysis on that instead of decoding one instruction at a time like the current setup.

Let me know what you guys think about the style of the line drawn and if you guys have any thoughts about improving it.

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -782,6 +782,59 @@ void QDisassemblyView::draw_function_markers(QPainter &painter, edb::address_t a
 }
 
 //------------------------------------------------------------------------------
+// Name: disassemble_at_address
+// Desc: Given a region and an address, returns the instruction at address.
+//------------------------------------------------------------------------------
+
+edb::Instruction QDisassemblyView::disassemble_at_address(edb::address_t address) {
+	IAnalyzer *const analyzer = edb::v1::analyzer();
+	quint8 buf[edb::Instruction::MAX_SIZE + 1];
+
+	// do the longest read we can while still not passing the region end
+	int buf_size = qMin<edb::address_t>((region_->end() - address), sizeof(buf));
+
+	// read in the bytes...
+	if(!edb::v1::get_instruction_bytes(address, buf, &buf_size)) {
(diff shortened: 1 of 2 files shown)


10110111:
(Didn't look at the code yet, just checked out the result.)
What makes me unsatisfied about how it looks is that it looks like a "bracket" instead of an arrow (in OllyDbg it's an arrow). I.e. it looks as if it just braces some block of code, not shows direction of a jump.
Another thing is that the up/down mini-arrows near current instruction seem redundant with this line. Maybe they should be hidden (but not for the other instructions).
Also, don't you feel that the line is too close to the code, namely it almost intersects with the up/down mini-arrows?

eteran:
I like this feature. We may need to add some horizontal padding to the left of disassembly in order to make drawing this look better. And also probably make the "arrow" a color that stands out a bit more.

10110111:
> And also probably make the "arrow" a color that stands out a bit more.
Note that if we look at OllyDbg, there red color means taken branch, and gray is non-taken one.

AaronOpfer:
I tried painting a unicode character over the end of the line to create an arrow:

image

but it looks ridiculous on very large font sizes:

image

eteran:
As I mentioned, I like the concept of this PR, but I'm not sure it's "ready" yet. It doesn't look quite right, particularly with regards to it looking more like a bracket than an arrow. 

So I'm going to close the PR, not because I don't want the feature, but because I think we can implement this in a way that results in a much more polished look.

If you are able to improve this though, please do resubmit it :-).

PR #485, opened 2016-12-01
Implement bare minimum PE32

AaronOpfer:
I'm not sure if this is enough to actually be able to run edb on windows, but this does successfully decode the header of a 32bit windows application running under wine on my machine.

I also removed a debug string I had added because it was never actually triggering before because every single region that should have had no binary at all was instead being picked up as PE32 since it had no sanity checks at all.


eteran:
nice! There is a ton that needs to be done to get windows building again though. The DebuggerCore API has been changed drastically. But one step at a time :-)

PR #489, opened 2016-12-02
Improve Binary String search performance

AaronOpfer:
This made finding a binary constant in my debuggee take about 5-10 seconds instead of not being complete after several minutes.

The main issue was the pointless usage of `ByteShiftArray` when we already had a contiguous view of the region memory. Constantly popping bytes into and out of the BSA destroyed all performance in my profiler (80% of the runtime). Calls to `region->size()` were second on the list (13% of runtime) because we called it inside the while loop.

### DIFF
--- a/plugins/BinarySearcher/DialogBinaryString.cpp
+++ b/plugins/BinarySearcher/DialogBinaryString.cpp
@@ -17,7 +17,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "DialogBinaryString.h"
-#include "ByteShiftArray.h"
 #include "edb.h"
 #include "IDebugger.h"
 #include "MemoryRegions.h"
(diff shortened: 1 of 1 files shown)


plugins/BinarySearcher/DialogBinaryString.cpp
```diff
@@ -101,7 +93,10 @@ void DialogBinaryString::do_find() {
 						}
 					}
 
-					ui->progressBar->setValue(util::percentage(i, regions.size(), p - &pages[0], region->size()));
+					// update progress bar every 64KB
+					if (quint64(p) & 0xFFFF == 0) {
```
10110111: This condition is never met.

AaronOpfer: I'm having trouble figuring out how to add compiler flags to a cmake project.

10110111: `cmake -DCMAKE_CXX_FLAGS=your-flags`

AaronOpfer: Okay, thanks. I'll make another PR addressing this and some other pedantic warnings I've littered in my PRs.

PR #493, opened 2016-12-05
Disassembly paint rewrite

AaronOpfer:
So I rewrote the disassembler paint event. My rewrite of the paint event code is somewhat opinionated but I think it might be better. Its structure lets us perform more high-level analysis when painting stuff (we can look at all of the disassembled instructions at once during one render pass) which might be useful if we decide to start painting more details someday.

To test my changes I inserted QElapsedTimer into the paintEvent and judged the elapsed time. My changeset doesn't result in an amazing speedup when analysis isn't working (maybe 6ms instead of 7ms? nothing amazing), but after analysis, in my massive debuggee binary, my paint times go from ~500ms to 6ms! I did this by adding a new method to the Analyzer which returns all functions that exist between a particular range of addresses, and then changing FunctionMap from a QHash into a QMap and then using funcs.lowerBound(startAddress-4096) to skip lots of irrelevant functions from iteration. This doesn't seem to have a significant impact on analysis times.

I also made some really minor tweaks which I can reverse if we don't want them:
 * I fixed several instances of strings not being clipped because we didn't use Qt's "width" argument to drawText correctly. such as...
 * Fixed address writing over line1; now will properly clip. (I did this before realizing @eteran made a workaround; I triggered it by changing font size in settings). Removed minimum size of line1 so that users an how hide addresses if they want to since they clip correctly now.
 * Added a red arrow icon for when EIP is on a breakpoint.
 * Render one extra line past the end of the window
 * Fixed the instruction bytes drawing so that it will no longer paint over line3 (clips correctly), and also slightly increased its minimum size so that it will always at least display ellipsis.
 * Add human-friendly string retrieval to edb.h

### DIFF
--- a/include/IAnalyzer.h
+++ b/include/IAnalyzer.h
@@ -32,7 +32,7 @@ public:
 	virtual ~IAnalyzer() {}
 
 public:
-	typedef QHash<edb::address_t, Function> FunctionMap;
+	typedef QMap<edb::address_t, Function> FunctionMap;
 
 public:
 	enum AddressCategory {
@@ -53,6 +53,7 @@ public:
 	virtual void analyze(const IRegion::pointer &region) = 0;
 	virtual void invalidate_analysis() = 0;
 	virtual void invalidate_analysis(const IRegion::pointer &region) = 0;
+	virtual bool for_funcs_in_range(const edb::address_t start, const edb::address_t end, std::function<bool(const Function*)> functor) const = 0;
 };
 
 #endif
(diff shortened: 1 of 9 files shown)


eteran:
Wow, this is great. Thanks!

Just out of curiosity. Why using `funcs.lowerBound(startAddress-4096) ` instead of using it to find `startAddress` exactly? And won't this fail for a particularly large function?

AaronOpfer:
If we use startAddress directly, we will not learn about functions whose
end addresses overlap with the startAddress. So I subtract 4KB instead.
Although it might make more sense now that I think about it to use start
Address without the arbitrary 4KB adjustment and then decrement the
iterator to search for these methods. That way we would not skip large
functions or iterate over 4KB of useless functions.

Issue #486, opened 2016-12-02
Register "badges" in the disassembly view.


eteran:
If we look at x86dbg, it has a nice (and simple) feature that I think edb would benefit from having too in some form:

![x86dbg](http://x64dbg.com/img/slide3.png)

If we look at the left side of the interface here, there is a visual notation that `RDX` happens to point to that location.  I would like us to have something similar, but am imagining something slightly more compact. Something like this (rough mock up)

![register_tag](https://cloud.githubusercontent.com/assets/571477/20821205/afb350ca-b80e-11e6-81b0-b5e70095580f.png)

That would optionally go immediately to the left of where the addresses are displayed.

AaronOpfer:
Where should we position them? We could render them overlaying the most significant digits of the address and not lose much visual information (since the significant digits are repeated above and below the address). Or we could add some blank space to the left of the address field for this purpose.

eteran:
I am imagining thrm going in some extra space to the left of the breakpoints

AaronOpfer:
I saw that there's a RegisterViewModelBase class but I'm not very familiar with Qt's modeling features. It looks very concerned about presentation layer which might make it inappropriate for directly interacting with it.

eteran:
Actually, if you get an instance of an `IState` object, it has a `gp_register(size_t n)` method which should get a register by index.

AaronOpfer:
Oh, I didn't see that `Register::name()` existed. Maybe this won't be so bad.

PR #497, opened 2016-12-08
Reg badges

AaronOpfer:
This adds register badges as described in #486 

 * I look for both registers and the values pointed to by registers. The latter might not be very useful but I don't think it hurts anything.
 * If multiple registers point to the same place we append the strings together (badge says `eax, ebx`. this will definitely make us paint over the addresses, though. 
 * They are always white text on blue background. We should probably figure out how we can use the palette to make these badges themeable instead.
 * I changed the line1/line2/line3 resize code so that if the user moves the mouse quickly, instead of ignoring the resize, we resize to the minimum.

### DIFF
--- a/include/Configuration.h
+++ b/include/Configuration.h
@@ -89,6 +89,7 @@ public:
 
 	// disassembly tab
 	Syntax            syntax;
+	bool show_register_badges;
 	bool 			  syntax_highlighting_enabled;
 	bool              zeros_are_filling;
 	bool              uppercase_disassembly;
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -116,6 +116,7 @@ void Configuration::read_settings() {
 	uppercase_disassembly = settings.value("disassembly.uppercase.enabled", false).value<bool>();
 	small_int_as_decimal  = settings.value("disassembly.small_int_as_decimal.enabled", false).value<bool>();
 	tab_between_mnemonic_and_operands=settings.value("disassembly.tab_between_mnemonic_and_operands.enabled", false).value<bool>();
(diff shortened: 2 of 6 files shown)


10110111:
My concern is that it's not possible to switch this off. As I've tried it, I usually just got some space eaten for nothing, since usually no registers have addresses visible on current page. Only sometimes these badges do appear, but again, IMO they are of not much use.

So I'd rather have an option to turn this off and reclaim the space.

AaronOpfer:
I agree that there probably should be an option to disable it. I'll take a look at how difficult that will be.

eteran:
Awesome! can you attach a screenshot of how it looks?

AaronOpfer:
image

image

eteran:
very cool. If you can make it toggleable as @10110111  suggested, I think it'll be ready to merge :-)

AaronOpfer:
I've added a config option. I've also squashed all the commits together.

The only thing I'm not happy about is that I can't listen to "config changed" events to determine if I can move l1, l2 and l3 to the left by 5 characters when users disable, and back again when users enable. I also have to hardcode the 5 character width in two different places, ideally I'd only have to hardcode it in one place. That's just how GUI code ends up going sometimes, though. It's not a big sin.

eteran:
I think that I created a configuration changed event at some point. Look to connect to `edb::v1::config()` `SIGNAL(settingsUpdated())`

eteran:
This looks good though, we can always improve it with further PRs. Merging!

eteran:
@10110111 BTW, just as an explanation of how I think this feature will be useful. I am thinking about how it could help in tracing self modifying code. Like self extracting archives and malware.

Either way, I am very happy with the state of this feature, as it can be toggled to the user's content. Customizing the UI is one of the strengths of edb :-D

PR #503, opened 2016-12-13
Remove "hinted" lookups from (I)Analyzer

AaronOpfer:
@eteran implemented a hint API on the Analyzer after I suggested it might make the analyzer function lookups faster, and it probably did. However, experience has taught us well, and now we know that `QMap::upperBound` could have been used instead as a much faster way to find the exact entry we were looking for thanks to the fact QMap uses skip lists. So I undid most of that work we did and replaced `find_containing_function`'s iteration with `upperBound` usage. I've tested for correctness but not performance, but I think we can all agree it is hard to imagine performance being any worse after this.

### DIFF
--- a/include/IAnalyzer.h
+++ b/include/IAnalyzer.h
@@ -44,12 +44,10 @@ public:
 
 public:
 	virtual AddressCategory category(edb::address_t address) const = 0;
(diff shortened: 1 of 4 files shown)


eteran:
Yea, i like the simpler API :-)

PR #510, opened 2016-12-15
Analysis improvements

AaronOpfer:
So it turns out that capstone-edb was string-formatting instructions in the `Instruction` constructor, even if the consumer of the instructions had no intention of ever using those strings for anything. By moving this string formatting into the Formatter (where it probably should have always been), we can improve performance for those use cases.

Notably, in my tests, this improves analysis performance by about *3.7x*.

```
[Analyzer] elapsed: 117828 ms
[Analyzer] elapsed: 31546 ms
```

### DIFF
--- a/plugins/Analyzer/Analyzer.cpp
+++ b/plugins/Analyzer/Analyzer.cpp
@@ -521,16 +521,6 @@ void Analyzer::collect_functions(Analyzer::RegionData *data) {
 		}
 	}
 
-	qDebug() << "----------Basic Blocks----------";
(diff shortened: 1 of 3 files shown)


eteran:
Will this do the adjustment multiple times if the instruction is "formatted" multiple times? Just want to get a gauge of if there are any cases that would cause a performance regression before merging.

AaronOpfer:
Yes it definitely would, you're right. However I do not see any place in
our code where we do that. We usually discard the results of disassembly
entirely and perform the whole process over again. If it becomes a problem
we can store a bunch of formatter state on the instruction I suppose.

PR #511, opened 2016-12-15
Remove redundant variable from Function

AaronOpfer:
We have a variable in `Function` that is only written to, never read from. This PR removes it.

@eteran what was the intended use of this variable? It's definitely not in use anymore. It also makes the constructor that takes an address dangerously misleading since the address isn't actually used for anything.

### DIFF
--- a/include/Function.h
+++ b/include/Function.h
@@ -83,7 +83,6 @@ public:
 	void swap(Function &other);
 
 private:
-	edb::address_t                   address_;
 	int                              reference_count_;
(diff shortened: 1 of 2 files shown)


eteran:
I imagine that it was the entry point of the function before I introduced the idea that functions are composed exclusively of basic blocks. And therefore the entry point is the first address of the first basic block.

I suspect the address based constructor is no longer used. If not, it should likely be factored out.

PR #513, opened 2016-12-16
Cache Syntax Highlighting results

AaronOpfer:
This improves rendering performance pretty significantly when scrolling around and clicking on stuff without jumping the viewport around a lot.

My paint times on my beefy machine went from 8-13ms to 2-3ms when the cache was getting hit consistently. This effectively eliminates a lot of the extra CPU cost of using syntax highlighting vs not using it. This reduced paint time also means that weaker machines will be more likely to keep up with 60fps (one frame every 16ms).


eteran:
So unfortunately, on my HiDPI screen with scaling, this results in terribly blurred rendering :-( I am going to revert this until we can come up with a better implementation:

edb_screenshot_20161218_022014

eteran:
It seems that if we rasterize too early (like to a `QPixmap` instead of a widget), it is done at a lower resolution and crappily upscaled when we render that to the widet.

So perhaps as a first pass of something that we know is safe, we can cache the results of: `SyntaxHighlighter::highlightBlock` as that is a loop of regex's and probably one of the more expensive parts of rendering?

PR #514, opened 2016-12-16
Cache analysis bar paint results

AaronOpfer:
When there are a HUGE number of functions the analysis bar can take up to 35ms to paint on my machine. Since the analysis bar looks exactly the same on basically every render, we instead paint the base analysis bar onto a buffer and cache it. Then, if none of the dimensions or number of functions have changed, we blit the buffer back and then paint the yellow triangle "You are here" indicator over the buffer.

This dramatically improves the performance of the Analyzer bar paint event which now takes 1ms. When combined with my other pull request, I finally get full frame rates while scrolling around in edb after my binary has been analyzed.

### DIFF
--- a/plugins/Analyzer/AnalyzerWidget.cpp
+++ b/plugins/Analyzer/AnalyzerWidget.cpp
@@ -66,42 +66,51 @@ void AnalyzerWidget::paintEvent(QPaintEvent *event) {
 
 	Q_UNUSED(event);
 
-	QPainter painter(this);
-	painter.fillRect(0, 0, width(), height(), QBrush(Qt::black));
-	QFontMetrics fm(font());
-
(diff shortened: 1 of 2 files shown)


10110111:
For the reference, what are your CPU and GPU?

AaronOpfer:
My main machine is an i7-4770 with an nvidia GTX 780.

So 35ms to paint is pretty high for the analysis widget. So, it just occurred to me that maybe instead of painting the analyzer by iterating over all of the functions, we should instead paint by iterating over all of the X pixels and then look up every one of those in the function map to determine what color that vertical bar at that x pixel should be colored. That way our render time doesn't depend on how large our binary is but instead how wide the window is, which might be faster. but this cache solution seems to work really well, so it probably is not necessary.

PR #517, opened 2016-12-18
Fix segfault in Analyzer Widget

AaronOpfer:
Forgot to initialize the `QPixmap* cache_` polnter to `nullptr`. It must not have crashed on my system because of a different memory allocator, but when I tried on my laptop running an older debian it crashed quickly.


eteran:
Does this possibly address 516?

AaronOpfer:
His backtrace doesn't look like mine which is squarely inside the analyzer
widget, his is inside the ollydbg register widget.

PR #518, opened 2016-12-18
Cache syntax highlighting

AaronOpfer:
@eteran can you try this version? It looks like all we had to do extra was copy the `devicePixelRatio` of the parent widget.

It would be nice if there was a way to automatically copy these attributes of the parent widget...

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -698,6 +698,7 @@ int QDisassemblyView::draw_instruction(QPainter &painter, const edb::Instruction
 					textLayout.endLayout();
 
 					map = new QPixmap(opcode.length() * font_width_, line_height);
+					map->setDevicePixelRatio(devicePixelRatio());
 					map->fill(Qt::transparent);
 					QPainter cache_painter(map);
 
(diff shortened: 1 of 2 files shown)


eteran:
This makes sense, I'll give it a try sometime tonight and will merge if it works as expected :-)

AaronOpfer:
Oh, `devicePixelRatio` doesn't exist on qt4. Let me `ifdef` that.

eteran:
Looked great on Qt4, but was unusable on my Qt5 build :-(.

edb_screenshot_20161218_191757

eteran:
BTW, I think as another layer of caching, we could have the `SyntaxHighlighter` itself cache `QVector<QTextLayout::FormatRange>` it produces for a given input string. So if the same instruction is rendered multiple times it won't have to process it at all :-)

AaronOpfer:
I think we can make the pixmap cache work, but the ifdefs would be a little
ugly. If a different cache is simpler and results in a nearly as good of a
speedup it might be worth it to shed the complexity. It probably also has a
faster cache miss scenario. I'll give it a try soon.

AaronOpfer:
So I tried caching the vector returned by SyntaxHighlighter but the results were not nearly as good as the pixmap version. My render times on my laptop were about 30ms without any caches, closer to 15ms with the cache on the syntax highlighter, and about 4ms with the pixmap. Although all these tests were done on debug builds.

AaronOpfer:
@eteran this last commit should fix it. I also found a DPI setting on my DE so I was able to test it with some confidence.

eteran:
Nice, doing some testing now. I suspected that the `QPixmap` would outperform the `QVector` caching. But I do wonder if there is any benefit to doing both. For example, the `SyntaxHighlighter` itself could have a cache in itself. And the `QDisassemblyView` could cache what was drawn.

Issue #414, opened 2016-05-13
Syntax-highlighted disassembly is vertically misplaced relative to address and instruction bytes


10110111:
See the screenshot. The syntax highlighted `vmulps` and `vsqrtss` instructions are rendered a pixel too high, while `nop`s are all correctly placed. The difference is that the former are syntax highlighed, while the latter are drawn as filling.
![edb-7px-font-syntax-highlighted-position-bug](https://cloud.githubusercontent.com/assets/6376882/15255431/9d908c14-1944-11e6-99da-f2843a70874d.png)
The above screenshot was made with disassembly view font _Droid Sans Mono 7_.
This bug will affect #337 once we want to apply its solution to other columns than just instruction bytes.

eteran:
So I'm having a hard time pinning this down :-/.

As an experiment, I made it render both using `draw_rich_text` and `painter.drawText` overlapping and as long as I ignore things which are bolded/italic it does look like it lines up perfectly on my machine (with that font).

eteran:
Looking at this screen shot more carefully , it *looks* to me like it depends on which line you are looking at! Look at the `nop` @ `001008a`, it has a small gap before the baseline. But the subsequent instruction doesn't look like it has this gap :-/. So this looks to me, like an issue with a non-integer character height or similar (I have a similar bug WAY back with a float based font width).

eteran:
Can you give the following patch a try and see if it helps at all? I've made it use `qreal` instead of `int` in some places int he hopes of catching these fractional font heights correctly:

https://gist.github.com/eteran/a1c46350c7e9cb4721e97b6850532bd2

10110111:
Doesn't seem to change anything. BTW, the text sometimes is more than one pixel displaced, see this screenshot taken on another machine, with _DejaVu Sans Mono 6_:
![screenshot - 180516 - 08 12 58](https://cloud.githubusercontent.com/assets/6376882/15348150/c6e25d06-1cd0-11e6-97d9-0b089e932e8c.png)
Here the colored text is two pixels above its correct place.

eteran:
hmm, unfortunately i am having a little difficulty reproducing the same effect so it's hard to know if i'm on the right track, but I'll keep plugging at it.

eteran:
PS. does it happen with larger fonts in any noticeable way? My previous difficulties with non-integer font dimensions were mostly visible when dealing with small fonts.

10110111:
[...]

PR #523, opened 2016-12-20
Fix font cache

AaronOpfer:
Related to #414. Try this change @eteran it works for me. Not sure why yours did not, though.


eteran:
works like a charm :-)

AaronOpfer:
I wish I knew why!

PR #533, opened 2017-01-15
New modern SVG Icons for debugger actions

akahan:


10110111:
Well it's not "new modern" — it's rather "taken from current theme", which might not be modern ;). Also, more than half of the patch is unrelated to icons at all. It should at least be split into two if you do intend to preserve both changes to accelerators and icons.

akahan:
Ok, no problem)

eteran:
@akahan Thanks for the initial PR. The difficulty with just using the theme for these icons in particular is that they are also used for the toolbar. So if for some reason they aren't there, the UI is broken. Some platforms (such as windows I believe) don't really have "icon themes" in Qt. So at least for these icons, I feel that we **must** include them in the source tree.

However, I will be looking at KDE's Breeze theme to see if their icons are higher quality than the ones we are currently using and consider if we should replace the current ones with the ones that KDE provided.

So please don't be discouraged, this PR was still helpful!

PR #534, opened 2017-01-16
SVG Icons for debugger actions

akahan:


10110111:
For some reason I have no icons after this patch, despite my Oxygen icon set does have them. Might be related to unexpected paths of installation but I'm not sure.

eteran:
@10110111 well, at the very least, this obviates the need for a proper fallback. I wonder if there is a way to specify it in the UIC file.

eteran:
@akahan Just wanted to keep you posted, I am not ignoring this PR, just have been a bit busy lately and will give things a proper look over soon :-).

eteran:
So, I think I am probably going to end up closing the PR, not because it is bad, but because it is incomplete for our needs.

Most of the these icons in particular are critical to the usability of the UI. So if they are not present, that's a "broken" setup. So, I think that we'll need to probably set important icons like this programatically so that we can specify a fallback icon of our choice. something like:

    QIcon::fromTheme("debug-run", ":/debugger/images/edb22-action-run.png");

I don't think we can do this in the xml files, if we can, i'd prefer that. Please re-open this if my assessment is wrong.

PR #543, opened 2017-03-12
Allow quoting variables in expressions

### LINKED ISSUE #400: Expression parser doesn't support variables with dashes in them
See issue #393 for more context. But in short when the user types:

`ld-2.22.so` the parser sees:

VAR: `ld`
OP: `-`
NUM: `2`
etc...

we need to figure out if there is a reasonable way to have the parser smart enough to work this out and determine that it is in fact a single name being specified.

AaronOpfer:
This change makes quoted expressions tokenized as a single variable.

Fixes #400.

Note that this does not make any changes to the expression dialog's auto-completion. After activating an entry in its auto-completer it will immediately say "Invalid Expression" afterward. I wanted to fix this, but I don't see any good ways. Adding quotes around all symbols that have spaces will make them harder to find with the auto-completer. Ideally we would quote variables as they are inserted, but I don't see any clear and obvious correct way of doing that.

### DIFF
--- a/include/Expression.tcc
+++ b/include/Expression.tcc
@@ -521,6 +521,22 @@ void Expression<T>::get_token() {
 				token_.set(">", Token::GT, Token::OPERATOR);
 			}
 			break;
+		case '"':
+			++expression_ptr_;
+			// Begin a quoted string
+			{
+				QString temp_string;
+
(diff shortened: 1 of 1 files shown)


eteran:
Fantastic!

PR #554, opened 2017-05-06
Resize 0->l1 area size to contain address badge

### LINKED ISSUE #552: Registers pointing to visible address may take more space than allocated in the Disassembly View
And I can't really move the address column. See this screenshot:

scrn

AaronOpfer:
`line1`, `line2` and `line3` indicators are now relative to the new `line0` which represents space allocated on the leftmost part of the widget rendering area for various extras (for now, just the register badges).

Fixes #552.

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -905,10 +905,6 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 		partial_last_line_ = false;
 	}
 
-	const int l1 = line1();
-	const int l2 = line2();
-	const int l3 = line3();
-
 	if(!region_) {
 		return;
 	}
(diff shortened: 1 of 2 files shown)


10110111:
Hmm, it does seem a bit jumpy now... But I guess it's tolerable. At least I don't know how it could be improved.

AaronOpfer:
Yeah, I'm not super pleased with the tendency of the contents to "jump" when items scroll in and out of view. But I was thinking we could use this same mechanism for the relative-jump line drawing later on.

10110111:
IMO, relative-jump line should be closer to the disassembly, like the up/down arrows are. And it would be really bad if it also made the columns jump. In OllyDbg space for this line is always reserved, and I think it's OK since it doesn't take more than a few pixels.

AaronOpfer:
Yeah, in that case judge this PR on its merit than compared to what other features we might use "line0" for.

eteran:
I have a thought about this, but it may be hard to describe. Basically, I am imagining that if enabled, that the space would always occupy the size of a single register in the tag, but with a little bit of slack for an ellipsis... But if you mouse over a tag with an ellipsis, I imagine the tag just being draw wider **over** the adjacent content so long as your mouse is within the tag. Or something to that effect.

When extended, it would be on top of any lines or adjacent text, and similar to a tooltip, but simpler and more under our control.

It's late so this may be a terrible idea ;-). What are you guy's thoughts?

eteran:
I think this patch looks good, there is of course room for improvement with the jumpiness, but perfect can be the enemy of good.

PR #569, opened 2017-07-17
Add a list of packages to install for the dependencies

a-r-d:
This should help people get all of the dependencies installed on a common distro such as Ubuntu.


eteran:
The wiki already has this information though...

https://github.com/eteran/edb-debugger/wiki/Compiling-(Ubuntu-Debian)

a-r-d:
Oh whoops my mistake, maybe readme can link to the wiki, because I had no idea it was there.

eteran:
No problem, I appreciate the contribution anyway :-).

I think a link to the wiki is probably a good idea.

PR #585, opened 2017-08-09
Some UX improvements

### LINKED ISSUE #347: How to switch to another thread?
I've tried double-clicking the row I'm interested in in Threads dialog, but it doesn't seem to work. How is the user supposed to switch between threads?

10110111:
This makes some usability improvement in DialogAttach as well as DialogThreads. Now it's possible to switch current thread. The latter improvement fixes #347.

### DIFF
--- a/include/IProcess.h
+++ b/include/IProcess.h
@@ -57,6 +57,7 @@ public:
 	// only legal to call when attached
 	virtual QList<std::shared_ptr<IThread>>  threads() const = 0;
 	virtual std::shared_ptr<IThread>         current_thread() const = 0;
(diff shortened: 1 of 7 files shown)


eteran:
Also, I believe given that we know due to design decisions that the `IThread` will always be an instance of `PlatformThread`, we can get away with a `static_cast` safely.

10110111:
I pass it by reference to make it clear that by contract the function expects a valid Thread object. As for why it's non-const, I decided to do this because in general an implementation may want to mutate the thread object to mark as used or whatever.
About `static_cast` though, I do agree.

10110111:
I've amended the commit to use `static_cast`.

PR #589, opened 2017-08-12
Some work on ARM disassembly syntax highlighting + Thumb support

10110111:
Also added ability to step over and fixed _Follow_ functionality. Single-stepping is still unsupported, but step over+Follow+RunTo is better than nothing :)
Also made ARM Assembler plugin work, currently only for ARM mode (i.e. not Thumb).


eteran:
This is fantastic work, thanks!

PR #591, opened 2017-08-20
ODbgRegisterView@ARM32: GPRs+CPSR

10110111:
This is an implementation of the register view, which supports GPRs and CPSR. I'm not too happy with how the `GE` and `IT` fields look, but for now I think it's good enough — at least better than without them.


eteran:
This is fantastic! Thanks so much for the effort.

PR #593, opened 2017-08-26
Basic single-stepping for ARM + various minor improvements

10110111:


plugins/DebuggerCore/unix/linux/PlatformThread.cpp
```diff
@@ -87,6 +87,7 @@ int PlatformThread::priority() const  {
 // Desc:
 //------------------------------------------------------------------------------
 edb::address_t PlatformThread::instruction_pointer() const  {
+	// FIXME(ARM): doesn't work at least on ARM32
```
eteran: On my kernel version this is unreliable for x86-64 as well. I don't know which version broke this, but it also seemed not super critical at the time since I only noticed it being used by the thread state display (not for things like the register view and such).

PR #604, opened 2017-09-14
EDB on Windows with MSVC

mistersandman:
This enables building EDB without plugins on Windows with CMake and Visual Studio 2017. Moreover, a CI configuration for AppVeyor is provided.

Pull requests for building the plugins will follow.


cmake/Modules/FindCapstone.cmake
```diff
@@ -0,0 +1,29 @@
+if (UNIX)
+  find_package(PkgConfig REQUIRED)
+
+  pkg_check_modules(CAPSTONE REQUIRED capstone>=3.0.4)
+  include_directories(${CAPSTONE_INCLUDE_DIRS})
```
10110111: Are this line and the next one actually needed? It seems they are present in the calling `CMakeLists.txt` anyway (after `find_package(Capstone REQUIRED)`), regardless of the platform.

mistersandman: No, you're perfectly right, they don't belong there. Already fixed, thank you!

mistersandman:
Unfortunately, moving the using declaration out of its anonymous namespace does not help resolve the ambiguity. It might even be a bug in Visual Studios argument-dependent lookup, but I didn't bother investigating that further.

mistersandman:
I took a closer look at this `make_unique` ambiguity. As long as the function call is unqualified, Visual Studio 2017 will always find `util::make_unique` via unqualified name lookup and `std::make_unique` via argument-dependent lookup which is the correct behavior, but results in an ambiguity.

Instead of qualifying all the affected function calls like I did in d011dd2, we could also simply exclude the definition of `util::make_unique` from all affected Visual Studio versions via #ifdefs.

10110111:
I think not exclude, but rather set to something like `template<typename...Ts> using make_unique=std::make_unique<Ts...>;` in place of our custom definition for VS.

mistersandman:
Something like this?
`#ifdef _MSC_VER
	using std::make_unique;
#else
	using util::make_unique;
#endif`

10110111:
Not quite, something like this, so that any references to `util::make_unique` in other files would also work:
```
diff --git a/include/Util.h b/include/Util.h
index b727a30..bd5490c 100644
--- a/include/Util.h
+++ b/include/Util.h
@@ -52,12 +52,16 @@ public:
    operator ConcreteEnum() const { return static_cast<ConcreteEnum>(data); }
 };
 
+#ifdef _MSC_VER
+using std::make_unique;
+#else
 // Until EDB switches to C++14, this will be useful
 template<typename T, typename ...Args>
 std::unique_ptr<T> make_unique( Args&& ...args )
 {
     return std::unique_ptr<T>( new T( std::forward<Args>(args)... ) );
 }
+#endif
 
 template<typename T>
 typename std::make_unsigned<T>::type to_unsigned(T x) { return x; }
```

mistersandman:
Ah, I see, works like a charm! Done.

eteran:
This is a FANTASTIC start! Thank you so much, I'm going to spend a day or two looking over things, but I plan on merging by the end of the weekend. :-)

10110111:
Which commit/line are you commenting on? GitHub doesn't seem to provide any context...

mistersandman:
The RunUntilRet symbol in the instantiation in Debugger::on_actionRun_Until_Return_triggered() (src/Debugger.cpp line 2735) is ambiguous. It can be the class <anonymous namespace>::RunUntilRet declared and defined above (src/Debugger.cpp line 163) or the friend declaration of RunUntilRet in the declaration of class Debugger (src/Debugger.h line 63). The problem is, that at the friend declaration in class Debugger, the class symbol RunUntilRet is not known yet and therefore the friend declaration automatically forward declares RunUntilRet, however without the anonymous namespace, thus resulting in two different symbols. By forward declaring RunUntilRet in an anonymous namespace, the friend declaration refers to this symbol instead of forward declaring its own symbol, thus removing the ambiguity.

As far as I understood, all anonymous namespaces of a translation unit are merged into a single anonymous namespace. Therefore, you would indeed have two unrelated declarations of `class A` if the two declarations appear in two different translation units. I *think* this should not pose a problem here, as the forward declaration is only used for the friend declaration, which in turn is only relevant for the RunUntilRet class defined in src/Debugger.cpp. However, this is certainly not an optimal design.

As far as I can see, the friend declaration is not even needed. So maybe we could get away by simply removing it.

10110111:
> As far as I understood, all anonymous namespaces of a translation unit are merged into a single anonymous namespace.
Right, relevant citation from C++2011 Standard, 7.3.1.1:

> An _unnamed-namespace-definition_ behaves as if it were replaced by
> ```
> inline namespace unique { /* empty body */ }
> using namespace unique ;
> namespace unique { namespace-body }

> where `inline` appears if and only if it appears in the _unnamed-namespace-definition_, all occurrences of _`unique`_ in a translation unit are replaced by the same identifier, and this identifier differs from all other identifiers in the entire program...

eteran:
I stand corrected :-).

These patches look like a great start to getting win32 support! merging :-)

mistersandman:
Thanks a lot for merging!

Note: You might want to integrate AppVeyor (https://www.appveyor.com/) with your GitHub project for CI builds on Windows. I included a build configuration that should work (appveyor.yml). It almost works the same as Travis CI and it is also free for open source projects.

Issue #134, opened 2015-03-29
DEFAULT_PLUGIN_PATH shouldn't be stringified by two levels of macro


GoogleCodeExporter:
```
If DEFAULT_PLUGIN_PATH contains other macro names, 
TOSTRING(DEFAULT_PLUGIN_PATH) at 
https://code.google.com/p/edb-debugger/source/browse/trunk/src/Configuration.cpp
#60 will produce wrong result.

For example，DEFAULT_PLUGIN_PATH=/usr/lib/x86_64-linux-gnu/edb, 
TOSTRING(DEFAULT_PLUGIN_PATH) will produce "/usr/lib/x86_64-1-gnu/edb", not 
"/usr/lib/x86_64-linux-gnu/edb".
This is because the word "linux" is defined at another place:
$ cpp -dM < /dev/null | grep linux
#define __linux 1
#define __linux__ 1
#define __gnu_linux__ 1
#define linux 1

```

Original issue reported on code.google.com by `xuzhen...@gmail.com` on 2 Aug 2014 at 8:03

PR #608, opened 2017-09-27
Fix default plugin path

sbrun:
The default plugin path is invalid when it embeds other macros like i386.
This was already reported in #134.

To avoid this problem, we define the DEFAULT_PLUGIN_PATH macro as a quoted string ready to be used in the code (the STRINGIFY trick is thus no-longer needed).

Kali bug: https://bugs.kali.org/view.php?id=4265


eteran:
Thanks! This looks reasonable enough of a fix.

Issue #527, opened 2016-12-20
Preserve comments through reboot


AaronOpfer:
Comments let a user place notes on some instructions which can be useful when debugging. They are not preserved through a reboot of edb or a restart of the debuggee which limits their usefulness. These should instead be saved.

eteran:
The bookmark plugin now demonstrates storing this type of information, should be pretty easy to add some stuff for the `QDisassemblyView`

RabidCicada:
I can take this on and see if I can get this in:)

eteran:
@RabidCicada So I think that the main thing we need to worry about here is comments in the disassembly view. Since this isn't a plugin, the hooks for save/restore sessions isn't auto-magical. The functions that you are likely interested in are found in Debugger.cpp:

    void Debugger::save_session(const QString &session_file);
    void Debugger::load_session(const QString &session_file);

Some notes:

* Best to store any addresses as hex strings, because 64-bit values round trip through JSON "numbers" are lossy for some values since JSON numbers are technically doubles for a fully conformant implementation.
* May be best to store the module ("region") base with the address, so we can (eventually) deal with ASLR on reloads.

AaronOpfer:
So I just tested this and comment saving only works when ASLR is disabled. Seems like this needs a little more work.

eteran:
@AaronOpfer definitely ASLR sensitive.

To deal with ASLR we'll need to start comments as region+offset instead of absolute addresses. Should be doable, and at the very least is easier to deal with than breakpoints.

eteran:
Implemented! (and ASLR aware)

PR #612, opened 2017-09-30
SessionManager implemented with comment saving function #527

helvete003:
I started to work on #527 and made a first (very basic) version that saves the comments in the session file and restores them when attaching the binary again. Before going any further with it i would like to hear from you if the code is even remotely good (my first ever contribution to any project)

### DIFF
--- a/include/Types.h
+++ b/include/Types.h
@@ -370,5 +370,11 @@ std::ostream&>::type operator<<(std::ostream& os, T val)
 	return os;
 }
 
+/* Comment Type */
+typedef struct {
+	edb::address_t 	address;
+	QString        	comment;
+} Comment;
+
 #include "ArchTypes.h"
 #endif
--- a/include/Types.h
+++ b/include/Types.h
@@ -371,10 +371,10 @@ std::ostream&>::type operator<<(std::ostream& os, T val)
 }
 
 /* Comment Type */
-typedef struct {
+struct Comment {
 	edb::address_t 	address;
 	QString        	comment;
-} Comment;
+};
 
 #include "ArchTypes.h"
 #endif
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,6 +53,7 @@ include_directories(
 	"qhexview"
 	"widgets"
 	"qjson4"
+	"session"
 )
 
 set(edb_SRCS
@@ -94,6 +95,8 @@ set(edb_SRCS
 	RegisterViewModelBase.cpp
 	State.cpp
 	SymbolManager.cpp
+	session/SessionManager.cpp
+	session/SessionError.cpp
 	ThreadsModel.cpp
 	widgets/LineEdit.cpp
 	widgets/NavigationHistory.cpp
@@ -153,7 +156,7 @@ set(edb_INCLUDES
 	${PROJECT_SOURCE_DIR}/include/QLongValidator.h
 	${PROJECT_SOURCE_DIR}/include/QULongValidator.h
 	${PROJECT_SOURCE_DIR}/include/HexStringValidator.h
-	${PROJECT_SOURCE_DIR}/include/Instruction.h		
+	${PROJECT_SOURCE_DIR}/include/Instruction.h
 	qjson4/QJsonArray.h
 	qjson4/QJsonDocument.h
 	qjson4/QJsonObject.h
@@ -161,7 +164,7 @@ set(edb_INCLUDES
 	qjson4/QJsonParser.h
 	qjson4/QJsonRoot.h
 	qjson4/QJsonValue.h
-	qjson4/QJsonValueRef.h	
+	qjson4/QJsonValueRef.h
 )
 
 if((${CMAKE_SYSTEM_PROCESSOR} MATCHES "i[3456]86") OR (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64") OR (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64"))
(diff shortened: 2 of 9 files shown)


eteran:
I'll take a look over the weekend. Thanks for your contribution!

include/Types.h
```diff
@@ -370,5 +370,11 @@ std::ostream&>::type operator<<(std::ostream& os, T val)
 	return os;
 }
 
+/* Comment Type */
+typedef struct {
```
10110111: Why should this be a `typedef` for an unnamed struct? Why not simply `struct Comment {`?

src/CMakeLists.txt
```diff
@@ -153,23 +156,23 @@ set(edb_INCLUDES
 	${PROJECT_SOURCE_DIR}/include/QLongValidator.h
 	${PROJECT_SOURCE_DIR}/include/QULongValidator.h
 	${PROJECT_SOURCE_DIR}/include/HexStringValidator.h
-	${PROJECT_SOURCE_DIR}/include/Instruction.h
```
10110111: Such whitespace changes shouldn't go into the commit if they are unrelated to commit's objective: they distract the reader from the "useful" changes.

src/Debugger.cpp
```diff
@@ -2873,7 +2870,18 @@ void Debugger::set_initial_debugger_state() {
 
 	const QString filename = session_filename();
 	if(!filename.isEmpty()) {
-		load_session(filename);
+		SessionError session_error;
+		if(!SessionManager::instance().load_session(filename, session_error)) {
+			QMessageBox::warning(
+				this,
+				tr("Error Loading Session"),
+				tr(session_error.getErrorMessage().toUtf8().constData())
```
10110111: 1. Why the `.toUtf8().constData()`? `QObject:tr` already takes a `QString` as input, these conversions are superfluous, or am I missing something?
2. Qt Linguist won't find the string to give to translators, so this call to `tr` will never work (but see `QT_TR_NOOP`).

src/session/SessionManager.cpp
```diff
@@ -0,0 +1,205 @@
...
+//------------------------------------------------------------------------------
+bool SessionManager::load_session(const QString &session_file, SessionError& session_error) {
+
+  QFile file(session_file);
+	if(file.open(QIODevice::ReadOnly | QIODevice::Text)) {
+		QByteArray json = file.readAll();
+		QJsonParseError error;
+		auto doc = QJsonDocument::fromJson(json, &error);
+		if(error.error != QJsonParseError::NoError) {
+      session_error.err = SessionError::UnknownError;
```
10110111: Indentation appears broken here.

include/Types.h
helvete003: Damn sorry, I am used to using struct with typedef so that i don't have to write struct in front of every parameter where i am using it but i forgot that i don't have to do this with c++ anymore.

src/Debugger.cpp
helvete003: Sorry, this wasn't my intention

src/CMakeLists.txt
helvete003: Sorry, this wasn't my intention

src/Debugger.cpp
helvete003: This was just a workaround to shut the compiler up. I was planing on using QT_TR_NOOP but had some trouble finding out if this would be the best way to got.

src/session/SessionManager.cpp
helvete003: Weird, this seems to be a problem with my IDE (everything is fine there) but VIM is showing the same indentation error... Sorry

src/Debugger.cpp
10110111: Ah, sorry, point 1) is indeed wrong, there's no such overload.

src/session/SessionManager.cpp
10110111: It's due to tab size. On github it's 8 chars by default.

10110111: This just means one should be consistent in using tabs vs spaces for indentation (in EDB the syle is to prefer tabs).

helvete003: Looked into the settings of my IDE and the tab size is 2 spaces. The indentation comes from my IDE trying to figure out the tab size of the moved code. Should i change the size in my IDE (found nothing in the style guide)?

10110111: Ideally, moved code should be moved verbatim, including any whitespace. It might be impossible to accomplish depending on the editor (although Vim should be able to do it, I've never had any problem with its yank-put functions).
As for setting up the IDE, I suppose you want to set tab policy to "never expand" and tab size for display to 4 chars (for Vim it's `set noexpandtab` and `set ts=4` respectively).

eteran:
Things are looking not bad! So, I also reviewed the patch, and certainly it is functional, and I think that we'll merge it in because this is a long wanted feature because this is a great start.  However, I plan to make some changes afterwards. A few notes:

1. I agree with all of @10110111 's thoughts, glad you made some appropriate changes.
2. an underscore followed by a capital is reserved in c++, I think I recall one variable having that in its name. For members, we prefer trailing underscores.
3. From a design point of view, instead of **duplicating** the comment data, and having the session manager aware of what comments are, I'd prefer that we make it so any object can register itself to be asked what it would like to save/restore by the manager and it just hand it a `QVariantMap` to save/load as JSON. This is more flexible and doesn't require the session manager to "know" about comments.
4. A singleton is "OK" for now, but we can do better from an API point of view. It's not tremendously different, but in edb.cpp, the model that we've been using is globals protected by namespace and scope, which are accessed by functions in `edb::v1` namespace. This is the "official API" of edb.

I'll probably merge either tomorrow or the day after (it's getting late :-P). And make some of the "fix-ups" shortly thereafter.

Once that is done, we can start to consider how to handle things like dynamically loaded memory regions :-).

Thanks again for the PR

eteran:
Merging now, I will make some tweaks to the code in the next couple of days. Thanks for the work!

PR #615, opened 2017-10-01
Mem Regions window: dblclick shows in CPU if --x

AaronOpfer:
If a region appears to be a code region, show it in the CPU view when double-clicked instead of the data-dump view.


AaronOpfer:
I'm wondering if a double-click action should also close the memory region dialog since the user probably doesn't want to look at it anymore: as a user it is difficult to tell that anything changed behind the dialog from my double-click action causing me to doubleclick again before realizing that the memory dump changed and I opened several of them.

PR #620, opened 2017-10-04
Deal with long double in MSVC

mistersandman:
Windows does not support 80bit long doubles natively. Instead a long double is equivalent to the 64bit double. Therefore, the conversion from edb::value80 to a long double in MSVC is not a trivial memcopy. Instead we leverage the FPU directly to load the 80 bits from edb::value80 and store them again in a double, doing the conversion for us.

Note that of course this reduces the precision and for values that can not be represented by a 64bit double it does not work (i.e. exponent too large -> inf, exponent too small -> 0). In order to fully support 80bit doubles, we would need to port a version of sprintf that can deal with these long doubles.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ before_build:
   - cmd: cd C:\projects
   - cmd: md build
   - cmd: cd build
-  - cmd: cmake -Wno-dev -G "%CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%configuration% -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DCAPSTONE_SDK="%CAPSTONE_SDK%" -DQt5Core_DIR="%QT_BASEDIR%\lib\cmake\Qt5Core" -DQt5_DIR="%QT_BASEDIR%\lib\cmake\Qt5" ..\edb-debugger
+  - cmd: cmake -Wno-dev -G "%CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%configuration% -DCMAKE_INSTALL_PREFIX=C:\projects\install -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DCAPSTONE_SDK="%CAPSTONE_SDK%" -DQt5Core_DIR="%QT_BASEDIR%\lib\cmake\Qt5Core" -DQt5_DIR="%QT_BASEDIR%\lib\cmake\Qt5" ..\edb-debugger
 
 build_script:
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -33,7 +33,7 @@ before_build:
   - cmd: cd C:\projects
   - cmd: md build
   - cmd: cd build
-  - cmd: cmake -Wno-dev -G "%CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%configuration% -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DCAPSTONE_SDK="%CAPSTONE_SDK%" -DQt5Core_DIR="%QT_BASEDIR%\lib\cmake\Qt5Core" -DQt5_DIR="%QT_BASEDIR%\lib\cmake\Qt5" ..\edb-debugger
+  - cmd: cmake -Wno-dev -G "%CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%configuration% -DCMAKE_INSTALL_PREFIX=C:\projects\install -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DCAPSTONE_SDK="%CAPSTONE_SDK%" -DQt5Core_DIR="%QT_BASEDIR%\lib\cmake\Qt5Core" -DQt5_DIR="%QT_BASEDIR%\lib\cmake\Qt5" ..\edb-debugger
```
10110111: Might it then be better to use inline asm instead of putting it into a separate file? I suspect we'll get lots of bug reports about broken builds if people use prefixes with whitespace.

mistersandman: Unfortunately, Visual Studio does not support inline assemly (at least in x64). Maybe we can check the CMAKE_INSTALL_REFIX when running cmake and warn about it, if we see that it contains whitespaces.

eteran: Yea, we should definitely look into a solution that allows whitespace (and if that's literally impossible, then at least warn the user at cmake time).

mistersandman: The problem is, that Visual Studio does escape quotation marks in preprocessor defines when calling cl.exe (the C/C++ compiler), but it does not escape them when calling ml64.exe (the assembler). DEFAULT_PLUGIN_PATH is composed using the CMAKE_INSTALL_PREFIX and is correctly wrapped in quotes, since it may contain whitespaces. But no matter how one manually escapes the quotation marks, either cl.exe or ml64.exe fails. I haven‘t found any option in CMake to specify preprocessor defines for a specific tool (i.e. only cl.exe but not ml64.exe, since for ml64.exe we really don‘t need the definition of DEFAULT_PLUGIN_PATH). Maybe we should use CMAKE_CXX_FLAGS to add the DEFAULT_PLUGIN_PATH instead of using CMake‘s add_definitions command.

PR #649, opened 2018-03-11
Properly update font in register view when settings are changed

sorokin:
At the moment register view font is not updated when settings are changed. This commit fixes that.

Although the change works, could you please review it, because I'm not sure the author intended it to be updated this way.


eteran:
Yea, this seems pretty reasononable

PR #651, opened 2018-03-12
RFC: introduce DebugEventHandlers (fix #650)

### LINKED ISSUE #650: Erasing an element of std::deque while iterating over it
I have found an error in the logic of step out, but before developing a fix I would like to discuss the problem.

First let me explain what I'm trying to do. Recently I have learned that there is a debugger like ollydbg, but for Linux. And I was really excited about all this. On the other hand the person who told me about edb warned me that he had found the debugger very unstable.

As I liked the idea of having olly-like debugger for Linux. I decided to take a look at it.

I shoved "-fsanitize=address,undefined -D_GLIBCXX_DEBUG" into CXXFLAGS. These options are to enable debug checks inserted by compiler and standard library. After this I got a program that can not even start without printing some errors. This is the way I found #645 and #646.

Another problem I found was an unconditional crash on step out. The problem is the following:

1. execute_debug_event_handlers() is run.
2. execute_debug_event_handlers() is iterating over handlers and now stays on element corresponding RunUntilRet.
3. RunUntilRet::handle_event() is run, prints "Found ret; passing back to debugger" and calls "delete this".
4. RunUntilRet::~RunUntilRet() is called doing remove_debug_event_handler(this).
5. remove_debug_event_handler() erases the element corresponding to RunUntilRet. Erasing the element invalidates the iterator that is hold by execute_debug_event_handlers (!!!!!)
6. remove_debug_event_handler() is quit, RunUntilRet::~RunUntilRet() is quit, RunUntilRet::handle_event() is quit.
7. The execution is back to execute_debug_event_handlers(). The next operation it does is ++it incrementing invalid iterator.

The console log I see is the following:
```
"Scanning for terminator at 0x4001c7: found pop"
"Scanning for terminator at 0x4001c8: found pop"
"Scanning for terminator at 0x4001c9: found ret"
"Found terminator ret at 0x4001c9"
"Setting breakpoint at terminator 0x4001c9"
"Event at address 0x4001ca"
Trap breakpoint
Previous was an internal breakpoint.
"On our terminator at 0x4001c9"
Found ret; passing back to debugger
safe_iterator.h:298:
Error: attempt to increment a singular iterator.

Objects involved in the operation:
    iterator "this" @ 0x0x7ffc15c979e0 {
      type = __gnu_debug::_Safe_iterator<std::__cxx1998::_Deque_iterator<IDebugEventHandler*, IDebugEventHandler*&, IDebugEventHandler**>, std::__debug::deque<IDebugEventHandler*, std::allocator<IDebugEventHandler*> > > (mutable iterator);
      state = singular;
      references sequence with type 'std::__debug::deque<IDebugEventHandler*, std::allocator<IDebugEventHandler*> >' @ 0x0x1993a00
    }
```

The problem is that the current implementation doesn't support unsubscribing from the debug events from the inside of debug event handler.

I see 3 ways of fixing this problem.
1. Using Qt signals. They support unsubscribing a slot from the inside. This solution involves making IDebugEventHandler QObject which I don't really like.
2. Using Boost.Signals2. They support unsubscribing a slot from the inside too. The problem is that Boost.Signals2 doesn't support unsubscribing by a pointer to the original object. Instead subscribing returns a special object (called connection) that is then used to unsubscribing. So this involves changing interface of add_debug_event_handler to return boost::signals2::connection.
3. Roll out home-grown mini-signal like class. This will help preserving the current interface. I expect the class to be 200-300 lines long.

I would like to use the solution 3.  What do you think?

P. S. As there must be more bugs lurking around can I suggest adding something like
```
set(CMAKE_CXX_FLAGS_DEBUG "-fsanitize=address,undefined -D_GLIBCXX_DEBUG ${CMAKE_CXX_FLAGS_DEBUG}")
```
to the CMakeLists.txt and asking all developers to test the program in this configuration?

sorokin:
OK. Here is my proposed fix for #650.

This problem is solved by introducing an class called
DebugEventHandlers. It holds IDebugEventHandler* and allows removing
when execute_debug_event_handlers is being in progress.

Here is the copy of its description:

DebugEventHandlers is a container of IDebugEventHandler*. It
supports three operations:
1. add -- Adding IDebugEventHandler*
2. remove -- Removing IDebugEventHandler*
3. execute -- Iterating over added handlers and calling them

The reason why a simple standard container is not enough is
that handlers can be added/removed while the list is being
iterated on. Naive implement would cause accessing an invalid
iterator in this case.

In order to support handler removal during iteration this class
keeps an extra bool (inside_execute_) -- whether execution of
handlers is in progress. If it is true then the elements of list
are not erased immediately. Instead they are marked for pending
removal by zeroing their handler pointer.

In order to support handler adding during iteration this class
uses std::list. std::list allows inserting elements without
invalidating iterators to existing elements. Newly added handlers
are inserted into the beginning of the list, so they are not
called on the event that is being executed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -65,6 +65,7 @@ set(edb_SRCS
 	CommentServer.cpp
 	Configuration.cpp
 	DataViewInfo.cpp
+	DebugEventHandlers.cpp
 	Debugger.cpp
 	DialogAbout.cpp
 	DialogArguments.cpp
--- a/src/DebugEventHandlers.cpp
+++ b/src/DebugEventHandlers.cpp
@@ -1,0 +1,99 @@
+/*
+Copyright (C) 2018 Ivan Sorokin
+                   vanyacpp@gmail.com
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "DebugEventHandlers.h"
+#include "Util.h"
+#include "IDebugEventHandler.h"
+
+DebugEventHandlers::DebugEventHandlers()
+	: current_handler_(nullptr)
+{}
+
+DebugEventHandlers::~DebugEventHandlers() {
+	if (current_handler_)
+		EDB_PRINT_AND_DIE("can not destroy debug events container while executing events");
+
+	if (!handlers_.empty())
+		EDB_PRINT_AND_DIE("some debug event handlers weren't property removed");
+}
+
+void DebugEventHandlers::add(IDebugEventHandler *handler) {
+	if (!handler)
+		EDB_PRINT_AND_DIE("event handler can not be nullptr");
+
+	for (auto i = handlers_.begin(), end = handlers_.end(); i != end; ++i)
+		if (*i == handler)
+			EDB_PRINT_AND_DIE("the same event handler is added twice");
+
+	// DebugEventHandlers::add can be called inside DebugEventHandlers::execute.
+	// Let's insert the handler in front so it is not called on the
+	// event currently being executed
+	handlers_.push_front(handler);
+}
+
+void DebugEventHandlers::remove(IDebugEventHandler *handler) {
+	if (!handler)
+		EDB_PRINT_AND_DIE("event handler can not be nullptr");
+
+	auto i = handlers_.begin();
+	for (auto end = handlers_.end(); i != end; ++i)
+		if (*i == handler)
+			break;
+
+	if (i == handlers_.end())
+		EDB_PRINT_AND_DIE("removal of an event that is not present");
+
+	// during execution only deletion of the current handler is supported
+	if (current_handler_ && current_handler_ != handler)
+		EDB_PRINT_AND_DIE("removal of non-current event handler during execution");
+
+	handlers_.erase(i);
+}
+
+edb::EVENT_STATUS DebugEventHandlers::execute(const std::shared_ptr<IDebugEvent> &event) {
+	if (current_handler_)
+		EDB_PRINT_AND_DIE("recursive debug event execution is not allowed");
+
+	// if somehow no handler is run, then let's just assume we should stop...
+	edb::EVENT_STATUS status = edb::DEBUG_STOP;
+
+	try {
+		// loop through all of the handlers, stopping when one thinks that it handled
+		// the event
+		for (auto i = handlers_.begin(), end = handlers_.end(); i != end;) {
+			// increment before processing, so if it's deleted it's not a problem
+			current_handler_ = *i++;
+			status = current_handler_->handle_event(event);
+			if (status != edb::DEBUG_NEXT_HANDLER) {
+				break;
+			}
+		}
+	}
+	catch (...) {
+		// reset current_handler_ to nullptr even if an exception was thrown
+		current_handler_ = nullptr;
+		throw;
+	}
+	current_handler_ = nullptr;
+
+	// if this assert fails, the bottom handler (which is owned by us) did something terribly wrong :-/
+	if (status == edb::DEBUG_NEXT_HANDLER)
+		EDB_PRINT_AND_DIE("the last event handler returned DEBUG_NEXT_HANDLER");
+
+	return status;
+}
--- a/src/DebugEventHandlers.h
+++ b/src/DebugEventHandlers.h
@@ -1,0 +1,77 @@
+/*
+Copyright (C) 2018 Ivan Sorokin
+                   vanyacpp@gmail.com
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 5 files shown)


eteran:
So I've looked at the issue, which thanks to your very thorough report was very easy to verify 👍.

I think any of your solutions would work, but I believe there is a simpler one...

Make a copy of the container, and it right over that instead in the execute function. (This is Qts approach n their foreach macro). We can avoid the cost of the copy, by using one of Qt's containers since those are copy on write.

sorokin:
Thank you. Copying of a container is indeed commonly used practice, and is likely to fix the problem in this case. Also it is much easier to implement, unfortunately it doesn't work in the general case.

Consider we have two handlers A and B and B is located after A in the handler list. Consider that A removed B from the list during the iteration? As the iteration is performed on a copy, B is still in the copy and will be called. It means that a handler is called even after its removal. This is dangerous as often removal is done in destructor and a class can not normally handle events after its removal.

Although this example can look artificial at first, it easily occurs in the programs that use signals widely. The problem with copying is that not only it doesn't support this case, it doesn't allow diagnosing (having an assert) this case occured. The program will just silently fail.

Even if supporting this case is considered unneeded, having an assert on unsupported cases is a good thing. That is why I haven't used copying for fixing this bug.

eteran:
That's a reasonable point. I can't look at your PR in-depth just this moment. But I'll be sure to look at it as soon as possible, since this bug should be addressed as quickly as we can.

eteran:
Looking at your PR, this is good quality code, and I am convinced it will work. Before I decide to merge, I'd like to explore the different approaches because whenever possible, I usually prefer the simplest solution that gets the job done correctly :-).

I was thinking that if we combine the copy ref counted container approach with a simple check of to see if the element was removed, then we can have the best of both worlds, but have the code be much simpler. The only downside is that it **may** be a little less efficient, but to be honest, I only expect these containers to have an average size of 1-3 elements, so it may be a nearly immeasurable difference. Something like this:

	EVENT_STATUS execute_debug_event_handlers(const std::shared_ptr<IDebugEvent> &e) {

		// if somehow no handler is run, then let's just assume we should stop...
		EVENT_STATUS status = DEBUG_STOP;
	
		// a copy so we can safely mutate as we iterate, QVector<> is ref counted, so this is very cheap on average
		const QVector<IDebugEventHandler*> handlers = g_DebugEventHandlers;

		// loop through all of the handlers, stopping when one thinks that it handled
		// the event
		for(auto it = handlers.begin(); it != handlers.end(); ++it) {
		
			IDebugEventHandler *handler = *it;
		
			// it wasn't removed, so it makes sense to execute it
			if(g_DebugEventHandlers.contains(handler)) {
				status = handler->handle_event(e);
				if(status != DEBUG_NEXT_HANDLER) {
					break;
				}
			}
		}

		// if this assert fails, the bottom handler (which is owned by us) did something terribly wrong :-/
		Q_ASSERT(status != DEBUG_NEXT_HANDLER);
		return status;
	}

In the average case, I would expect this to have basically no cost. additions to the list during execution would obviously simply take effect during the next event.

Thoughts?

eteran:
Another option that occurs to me, which is even simpler is to do 3 things:

1. switch to std::list<> so iterators aren't invalidated by insertion
2. state that as a rule, a handler may only remove itself. deleting other handlers is UB.
3. make the execute function look like this:

	EVENT_STATUS execute_debug_event_handlers(const std::shared_ptr<IDebugEvent> &e) {

		// if somehow no handler is run, then let's just assume we should stop...
		EVENT_STATUS status = DEBUG_STOP;

		// loop through all of the handlers, stopping when one thinks that it handled
		// the event
		for(auto it = g_DebugEventHandlers.begin(); it != g_DebugEventHandlers.end(); ) {
	
			// increment before processing, so if it's deleted it's not a problem
			IDebugEventHandler *handler = *it++;
		
			status = handler->handle_event(e);
			if(status != DEBUG_NEXT_HANDLER) {
				break;
			}
		}

		// if this assert fails, the bottom handler (which is owned by us) did something terribly wrong :-/
		Q_ASSERT(status != DEBUG_NEXT_HANDLER);
		return status;
	}

Less robust in that it doesn't support deleting other handlers within a handler, but I'm not sure that is really the most "sane" of circumstances to want to support. I can't really think of a use case which deleting other handlers seems like a good idea ;-)

Anyway, I am curious of your thoughts on this too!

sorokin:
In my opinion both of your proposed solution will work. Choosing between these two I like the option 2 (incrementing iterator before deferencing) more, but it is just a personal preference, not the result of rational reasoning.

Speaking of option 1 the only downside I see is its computational complexity which, as you mentioned, doesn't matter if the list is short.

The only thing I would like to add is the debug checks (or asserts) that only the currently executed handler is deleted. These checks can be optional and be enabled only in debug builds. I believe having debug checks like this is important as it makes testing more productive and it simplifies debugging.

I implemented the second approach (with corresponding debug checks). Here you can see the commit: https://github.com/sorokin/edb-debugger/commit/9021493579b9ffa046f7b2f1387bd4d4cbb20d4f, if you like it I can move it into this PR.

eteran:
I think the code for option 2 that you've written looks great. My only last question is where is EDB_PRINT_AND_DIE defined? I don't see it anywhere in the github diff but i assume it must be in there somewhere.

10110111:
> My only last question is where is EDB_PRINT_AND_DIE defined?

It's a pre-existing define in `Util.h`.

eteran:
@10110111 LOL, somehow I completely forgot that was added!

sorokin:
Thank you! EDB_PRINT_AND_DIE was just a random macro I have found when I searched for something to abort the program. 

As the behavior of DebugEventHandlers has changed I would like to rewrite the comment for it. Unfortunately I can not do this right now. I'll update the PR later this evening.

sorokin:
I updated the PR. I added a comment for the class describing its behavior. Otherwise the code should be identical to what I linked before (option 2).

As English is not my native language any corrections to the comment and the commit message are welcome.

eteran:
merged, looks great. Comments read fine too :-).

I may change the blocks like this:

	auto i = handlers_.begin();
	for (auto end = handlers_.end(); i != end; ++i)
		if (*i == handler)
			break;

into code using `std::find`, but beyond that the code looks very good. Thanks!

PR #653, opened 2018-03-13
Set consistent LC_NUMERIC

10110111:
One of my latest commits which changed the use of `std::istringstream` to parse floating-point numbers to `std::sto...` resulted in breakage of EDB in non-English locales. Namely, if `LC_NUMERIC` is set to something like `ru_RU.UTF-8`, decimal separator is comma. And since `std::stod` and family, unlike `std::istream`, is locale-aware (too aware, unfortunately), the numbers which passed our `FloatXValidator`, appear rejected in `fullStringToFloat`.
To fix this, EDB must set its `LC_NUMERIC` to `"C"`, which is done by this patch.

### DIFF
--- a/src/Debugger.cpp
+++ b/src/Debugger.cpp
@@ -476,6 +476,12 @@ Debugger::Debugger(QWidget *parent) : QMainWindow(parent),
 
 	// let the plugins setup their menus
 	finish_plugin_setup();
+
+	// Make sure number formatting and reading code behaves predictably when using standard C++ facilities.
+	// NOTE: this should only be done after the plugins have finished loading, since some dynamic libraries
+	// (e.g. libkdecore), which are indirectly loaded by the plugins, re-set locale to "" once again. (This
+	// first time is QApplication.)
+	std::setlocale(LC_NUMERIC, "C");
 }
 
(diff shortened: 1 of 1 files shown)


sorokin:
A minor correction: both `std::istringstream` and `stod` are locale-aware. The difference is that they are aware of different locales `std::istringstream` uses `std::locale` and stod uses C-locale (`setlocale`). See: https://habrahabr.ru/post/344108/ (the article is in Russian, sorry).

Good news: C++17 has a locale-independent way of converting numbers: `std::from_chars` (http://en.cppreference.com/w/cpp/utility/from_chars).

10110111:
Oh, there're two independent locales... Funny C++... Still, I suppose this patch remains correct, right?

As for `std::{from,to}_chars`, besides the fact that EDB uses only C++11, these functions are still not fully implemented even in GCC 8 (only integer overloads are implemented). In several years, though, they should become quite useful to print user-friendly floating-point values, so that we don't get `0.29999999999999999` for 64-bit `double` value `0.3` as we currently do with `max_digits10`.

eteran:
I agree that @10110111's patch still looks good.

PR #657, opened 2018-05-13
Make the desktop's entries somewhat better

Petross404:
Hello fellow contributors of edb. I made a small change in the `edb.desktop` file, if you like it consinder to accept it :)

As of now, the gentoo ebuild of edb-debugger uses `sed` to alter the description inside the `edb.desktop`, so I thought I could share my idea with a PR. This will help both me to clean up future ebuilds and perhaps other users of the software that want a more descriptive `.desktop` file.

Thanks!

### DIFF
--- a/edb.desktop
+++ b/edb.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=edb
-GenericName=edb debugger
-Comment=edb debugger 
+GenericName="Evan's Debugger"
+Comment="edb is a cross platform x86/x86-64 debugger"
 Exec=edb
 Icon=edb
 Terminal=false


edb.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=edb
-GenericName=edb debugger
-Comment=edb debugger 
+GenericName="Evan's Debugger"
+Comment="edb is a cross platform x86/x86-64 debugger"
```
10110111: Actually, ARM port is in progress. Namely, EDB already builds on AArch32 and even lets you do something. So it's not limited to x86/x86-64.

Petross404: If the "x86" part is your only concern, I can omit it :)

eteran: To be fair, the README and project description on github still say "x86/x86_64"... which I will address now :-).

edb.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=edb
-GenericName=edb debugger
-Comment=edb debugger 
+GenericName="Evan's Debugger"
```
mgorny: I'm sorry for not looking at this closer before telling you to submit your fixes. According to XDG spec, 'GenericName' is a generic description of applications class, so I think a plain 'Debugger' would be more correct. `Evan's Debugger` would be more correct for 'Name' key.

https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

eteran: No worries it's an easy fix

Issue #628, opened 2017-11-20
Spontaneous "Error Loading Session" occurring sometimes


10110111:
In some cases, I've not yet traced in which ones, when I run EDB as `edb --run /path/to/my/prog`, it opens its main window, and then shows me an empty warning message box, with title "Error Loading Session". Here's how it looks:
![screenshot](https://user-images.githubusercontent.com/6376882/33012045-c67a6a02-cdf0-11e7-8215-34d21d5ae2a7.png)
Stack trace from this state after I attach GDB to EDB:
```
0x00007efd55524ce0 in do_sigtimedwait (timeout=0x7ffc90cc4a90, info=0x7ffc90cc4b20, set=0x7ffc90cc4aa0) at ../nptl/sysdeps/unix/sysv/linux/../../../../../sysdeps/unix/sysv/linux/sigtimedwait.c:53
53      ../nptl/sysdeps/unix/sysv/linux/../../../../../sysdeps/unix/sysv/linux/sigtimedwait.c: No such file or directory.
(gdb) bt
#0  0x00007efd55524ce0 in do_sigtimedwait (timeout=0x7ffc90cc4a90, info=0x7ffc90cc4b20, set=0x7ffc90cc4aa0) at ../nptl/sysdeps/unix/sysv/linux/../../../../../sysdeps/unix/sysv/linux/sigtimedwait.c:53
#1  __GI___sigtimedwait (set=0x7ffc90cc4aa0, info=0x7ffc90cc4b20, timeout=0x7ffc90cc4a90) at ../nptl/sysdeps/unix/sysv/linux/../../../../../sysdeps/unix/sysv/linux/sigtimedwait.c:81
#2  0x00007efd42b0a019 in DebuggerCorePlugin::native::wait_for_sigchld (msecs=10) at /home/ruslan/mydev/edb-debugger/plugins/DebuggerCore/unix/DebuggerCoreUNIX.cpp:290
#3  0x00007efd42b0d103 in DebuggerCorePlugin::DebuggerCore::wait_debug_event (this=0x1aca560, msecs=10) at /home/ruslan/mydev/edb-debugger/plugins/DebuggerCore/unix/linux/DebuggerCore.cpp:588
#4  0x0000000000582774 in Debugger::next_debug_event (this=0x7ffc90cc5880) at /home/ruslan/mydev/edb-debugger/src/Debugger.cpp:3417
[...]

PR #696, opened 2019-02-18
show error message when session file fails to open

sorokin:
This is follow up of last comments in #628.

Now we show error messages when session file fails to open (for example in case permission are denied).


eteran:
Looks good, thanks!

PR #702, opened 2019-03-25
Address refactor

eteran:
I think I've got things working nicely in this branch :-) Time to merge it in :-)

Refactors edb::valueX types to not use inheritance and be simple for simple cases. This makes debugging easier for sure since it is easier to inspect the contents of address_t types in Qt-creator and **maybe** will encourage better code-gen (I'm not sure if it was ever bad, but a simpler wrapper around a basic type is at least as easy for the compiler to optimize if not better).

Also fixes a few UB issues regarding reading more from a variable than exists!

And finally, squashes a ton of warnings that popped up with gcc-8.2


include/Value.h
```diff
@@ -82,23 +82,18 @@ class value_type_large {
...
 		static_assert(sizeof(U) <= sizeof(T), "It doesn't make sense to expand a larger type into a smaller type");
 
+		value_type_large created;
+
 		auto dataStart = reinterpret_cast<const char*>(&data);
-		auto target    = reinterpret_cast<char*>(&value_);
+		auto target    = reinterpret_cast<char*>(&created.value_);
+
+		std::memcpy(target, dataStart, sizeof(U));
+		std::memset(target + sizeof(U), 0, sizeof(T) - sizeof(U));
```
10110111: It's better to refer to objects when doing offset calculations than to some (possibly unrelated) types. It'll 1) add more clarity (show why we're using this size and not another), and 2) will automatically update the sizeofs to refer to the same objects even if their types change.

include/Value.h
```diff
@@ -814,8 +828,8 @@ struct value_type80 {
 	}
 
 public:
-	bool operator==(const value_type80 &rhs) const { return memcmp(value_, rhs.value_, 10) == 0; }
-	bool operator!=(const value_type80 &rhs) const { return memcmp(value_, rhs.value_, 10) != 0; }
+	bool operator==(const value_type80 &rhs) const { return std::memcmp(value_, rhs.value_, sizeof(T)) == 0; }
```
10110111: This change looks scary. What if you later decide to make T equal in size to long double? You may then fail to notice that you're comparing padding bytes in 64-bit code.

plugins/DebuggerCore/unix/linux/arch/x86-generic/PlatformState.cpp
```diff
@@ -478,6 +481,9 @@ bool PlatformState::fillFrom(const X86XState &regs, size_t sizeFromKernel) {
 		// Only fill the registers which are actually supported, leave invalidity marks intact for other parts
 		if (avx.xcr0 & X86XState::FEATURE_AVX) { // If AVX state management has been enabled by the OS
 			for (size_t n = 0; n < MAX_YMM_REG_COUNT; ++n) {
+
+				// TODO(eteran): a default initialized value should just have all zeros set
+				// no need to use fromZeroExtended, right?
```
10110111: When I was writing all this stuff, I supposed that default values would be `bad1bad1bad1....` Zero is a very bad choice to default-initialize state data in EDB, since a bug in treating unfilled state/registers will then be hidden. And we have actually had such a bug — when I started making my first commits here.

include/Value.h
```diff
@@ -54,27 +54,25 @@ class large_value_type {
...
-		char buf[1024];
+		char buf[sizeof(T) * 2 + 1];
 		char *p = buf;
-		auto ptr = reinterpret_cast<const char *>(value_);
-		auto end = ptr + sizeof(T);
 
-		for(auto it = end; it != (ptr - 1); --it) {
-			p += sprintf(p, "%02x", *it & 0xff);
+		for(auto it = std::rbegin(value_); it != std::rend(value_); ++it) {
+			p += sprintf(p, "%016llx", *it & 0xffffffffffffffffULL);
```
10110111: Why do you need this `& 0xffffffffffffffffULL` now? It's a NOP. Besides, such a form of a number is not really readable: why not write e.g. `0xffff'ffff'ffff'ffffULL` taking advantage of C++14 digit separators feature? (or even `uint64_t(-1)`, to make the intent explicit)

include/Register.h
```diff
@@ -72,13 +72,7 @@ class EDB_EXPORT Register {
...
 
 	// Return the value, zero-extended to address_t to be usable in address calculations
-	edb::address_t valueAsAddress() const {
-		// This function only makes sense for GPRs
-		assert(bitSize_ <= 8 * sizeof(edb::address_t));
-		edb::address_t result(0LL);
-		std::memcpy(&result, &value_, bitSize_ / 8);
-		return result;
-	}
+	edb::address_t valueAsAddress() const;
```
10110111: This is a bit of a pessimization: the compiler is now unable to inline the (trivial!) function unless link-time optimization does this.

include/Value.h
eteran: I looked through your comments, and they generally all make sense :-). I'll implement some of the changes you suggest!

plugins/DebuggerCore/unix/linux/arch/x86-generic/PlatformState.cpp
eteran: Very good point. Glad that I made this a comment, and didn't just change it :-)

include/Value.h
eteran: it was left over from when I did it bytewise, you are right, isn't needed at all!

include/Register.h
eteran: Yea, on this one. I kinda tend to prefer to have the implementation in the source files unless it is something like a one-liner. If this ever shows up in a profile as being notably slow, well you can say you told me so ;-).

PR #709, opened 2019-05-08
Don't rely on /proc to get thread's instruction pointer

bet4it:
Linux after v4.9 has stopped reporting eip and esp in `/proc/PID/stat`
Refer: torvalds/linux@0a1eb2d474edfe75466be6b4677ad84e5e8ca3f5

Things need to do before merge this PR:
Implement this `TODO` in `get_state`:
https://github.com/eteran/edb-debugger/blob/fb7643f65927b769668e41ea20f008cdb82f7a19/plugins/DebuggerCore/unix/linux/arch/x86-generic/PlatformThread.cpp#L164-L165

### DIFF
--- a/plugins/DebuggerCore/unix/linux/DebuggerCore.cpp
+++ b/plugins/DebuggerCore/unix/linux/DebuggerCore.cpp
@@ -553,8 +553,10 @@ std::shared_ptr<IDebugEvent> DebuggerCore::handle_event(edb::tid_t tid, int stat
 		const auto signo = WSTOPSIG(status);
 		if(signo == SIGILL || signo == SIGSEGV) {
 			// no need to peekuser for SIGILL, but have to for SIGSEGV
+			State state;
+			(*it)->get_state(&state);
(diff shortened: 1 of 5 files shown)


plugins/DebuggerCore/unix/linux/DebuggerCore.cpp
```diff
@@ -553,8 +553,10 @@ std::shared_ptr<IDebugEvent> DebuggerCore::handle_event(edb::tid_t tid, int stat
 		const auto signo = WSTOPSIG(status);
 		if(signo == SIGILL || signo == SIGSEGV) {
 			// no need to peekuser for SIGILL, but have to for SIGSEGV
+			State state;
+			(*it)->get_state(&state);
```
10110111: This will fetch the whole state even when we don't need to (see the comment above).

eteran:
I think we can do things more efficiently than this PR. I appreciate the effort and apologize for letting it linger for so long. I'd rather alter the state API to be able to fetch specific classes of registers (or even specific important registers) without the need fetch the whole state unnecessarily.

PR #714, opened 2019-06-11
ARM regression fixes

10110111:
This fixes building of EDB on ARM32 for me. I don't know whether anything was broken for runtime, but at least compilation is fixed now.

Fixes most of the problems which are discussed in #713 (except the one with Capstone, which is the user-side problem, not EDB).


eteran:
So that static_assert that you made an assert... I don't see how it won't trigger at runtime, both operands are not dependant on runtime values, so it's either true or it isn't :-/.

I suspect that this will **compile** on arm, but if that code path is ever executed, it's probably a guaranteed assert.

10110111:
> I suspect that this will compile on arm, but if that code path is ever executed, it's probably a guaranteed assert.

The static assertion won't compile (because `long double` is 64-bit on ARM). The function is useless on ARM, but since we aren't trying to optimize for size, I didn't try to complicate the code to avoid this instantiation of the template.

eteran:
Gotcha, so it'll compile, but won't ever be run at runtime then. Makes sense.

PR #725, opened 2019-10-24
Draw jump lines on left-side of disassembly view

shahril96:
Added feature for drawing jump lines on the left side of disassembly view. Kinda missing this one when I was using EDB. (in Windows my goto debugger is x64dbg, so was inspired by that)

image

Any comment is really appreciated :) (as I have not majorly tested this one yet - only on x86-64 Linux).

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -58,6 +58,7 @@ struct WidgetState1 {
 	int line1;
 	int line2;
 	int line3;
+	int line4;
 };
 
 constexpr int default_byte_width = 8;
@@ -174,6 +175,7 @@ void QDisassemblyView::resetColumns() {
 	line1_ = 0;
 	line2_ = 0;
 	line3_ = 0;
+	line4_ = 0;
 	update();
 }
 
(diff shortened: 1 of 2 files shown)


eteran:
This looks.... fantastic! I'll have to look at the code, but the picture alone looks great.

Thanks!

eteran:
The code looks great, so I'm going to merge. I do however have one request. Could you make a followup PR to make this functionality optional? I can help with threading it through the config dialog. But you can look at the register tags feature as an example of how to do it.

10110111:
IMO, the chunk added in `QDisassemblyView::paintEvent` should have been made a separate method. `QDisassemblyView::paintEvent` has grown too large already.

@eteran In the future, could you please merges for a while so that watchers could also review PRs? Thanks.

eteran:
@10110111 of course, no problem. I agree that it could use a refactor, but to be honest I was a bit excited because I thought the result looked so good!

eteran:
@10110111 We probably should have a small helper function for each separate part of the rendering, so it's easier to manage.

10110111:
Fully agreen.

shahril96:
@eteran No problem :) I will do the followup PR, however, it is going to take a while (on holiday for the whole next week)

eteran:
@shahril96 no worries. I am currently doing the refactor that @10110111  and I spoke about, so no rush :-)

Thanks again!

PR #726, opened 2019-10-24
Improve jump arrows drawing

shahril96:
Previously, some jump arrows might be overlapped or being so close together (which makes it not to look nice).

This PR attempted to fix that, and now all jump arrows fit nicely.  :)

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -1228,6 +1228,8 @@ void QDisassemblyView::drawComments(QPainter &painter, const DrawingContext *ctx
 void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *ctx) {
 	std::vector<JumpArrow> jump_arrow_vec;
 
+	painter.setRenderHint(QPainter::Antialiasing, true);
+
 	for (int line = 0; line < ctx->lines_to_render; ++line) {
 
 		auto &&inst = instructions_[line];
@@ -1395,19 +1397,12 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 			painter.drawPolyline(points, 4);
 
 			// draw arrow tips
-			painter.setPen(QPen(arrow_color, arrow_width, Qt::SolidLine));
-			painter.drawLine(
-				end_x - (font_width_/3), 
-				dst_y - (font_height_/4), 
-				end_x, 
-				dst_y
-			);
-			painter.drawLine(
-				end_x - (font_width_/3), 
-				dst_y + (font_height_/4), 
-				end_x, 
-				dst_y
-			);
+			QPainterPath path;
+			path.moveTo(end_x, dst_y);
+			path.lineTo(end_x - (font_width_/2), dst_y - (font_height_/3));
+			path.lineTo(end_x - (font_width_/2), dst_y + (font_height_/3));
+			path.lineTo(end_x, dst_y);
+			painter.fillPath(path, QBrush(arrow_color));
 
 		} else if (is_dst_upward) {  // if dst out of viewport, and arrow facing upward
 
@@ -1420,19 +1415,12 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 			painter.drawPolyline(points, 3);
 
 			// draw arrow tips
-			painter.setPen(QPen(arrow_color, arrow_width, Qt::SolidLine));
-			painter.drawLine(
-				start_x - (font_width_/3), 
-				font_height_/4, 
-				start_x, 
-				0
-			);
-			painter.drawLine(
-				start_x + (font_width_/3), 
-				font_height_/4, 
-				start_x, 
-				0
-			);
+			QPainterPath path;
+			path.moveTo(start_x, 0);
+			path.lineTo(start_x - (font_width_/2), font_height_/3);
+			path.lineTo(start_x + (font_width_/2), font_height_/3);
+			path.lineTo(start_x, 0);
+			painter.fillPath(path, QBrush(arrow_color));
 
 		} else { // if dst out of viewport, and arrow facing downward
 
@@ -1445,21 +1433,16 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 			painter.drawPolyline(points, 3);
 
 			// draw arrow tips
-			painter.setPen(QPen(arrow_color, arrow_width, Qt::SolidLine));
-			painter.drawLine(
-				start_x - (font_width_/3), 
-				viewport()->height() - (font_height_/4), 
-				start_x, 
-				viewport()->height()
-			);
-			painter.drawLine(
-				start_x + (font_width_/3), 
-				viewport()->height() - (font_height_/4), 
-				start_x, 
-				viewport()->height()
-			);
+			QPainterPath path;
+			path.moveTo(start_x, viewport()->height());
+			path.lineTo(start_x - (font_width_/2), viewport()->height() - (font_height_/3));
+			path.lineTo(start_x + (font_width_/2), viewport()->height() - (font_height_/3));
+			path.lineTo(start_x, viewport()->height());
+			painter.fillPath(path, QBrush(arrow_color));
 		}
 	}
+
+	painter.setRenderHint(QPainter::Antialiasing, false);
 }
 
 //------------------------------------------------------------------------------
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -1339,9 +1339,8 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 	IProcess* process = edb::v1::debugger_core->process();
 	process->current_thread()->get_state(&state);
 
-	for (size_t i = 0; i < jump_arrow_vec.size(); i++) {
-
-		JumpArrow& jump_arrow = jump_arrow_vec[i];
+	for (const JumpArrow& jump_arrow : jump_arrow_vec) {
+		
 		bool is_dst_upward = jump_arrow.target < instructions_[jump_arrow.src_line].rva();
 		
 		// edges value in arrow line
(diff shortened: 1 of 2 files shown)


eteran:
Does this take into account my refactor that I push a little bit ago?

shahril96:
Yes, I've merged your commits (little while ago) with mine.

eteran:
OK, cool. I'll review it in a little bit and give @10110111 a chance to comment too! :-)

Thanks again, this stuff is looking great.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1287,7 +1287,7 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 		});
 
 		// find suitable arrow horizontal width
-		for (int i = 0; i < jump_arrow_vec.size(); i++) {
+		for (int i = 0; i < (int) jump_arrow_vec.size(); i++) {
```
eteran: Isn't `std::vector<>` indexed by an unsigned type like `size_t` or `std::vector<>::size_type`? It's hard to tell, but if `i` is never compared to a signed value, then we can probably just make it a `size_t` and not need this cast.

Also, even though it's more verbose, please prefer c++ style casts :-).

src/widgets/QDisassemblyView.cpp
```diff
@@ -1356,7 +1356,7 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 		IProcess* process = edb::v1::debugger_core->process();
 		process->current_thread()->get_state(&state);
 
-		for (int i = 0; i < jump_arrow_vec.size(); i++) {
+		for (int i = 0; i < (int) jump_arrow_vec.size(); i++) {
```
eteran: Same as above

shahril96:
You're welcome, happy to contribute a little bit. :)

src/widgets/QDisassemblyView.cpp
shahril96: Ah yes, you're correct. Fixed.

src/widgets/QDisassemblyView.cpp
shahril96: Fixed.

10110111:
Some comments on the whole changeset including the previous (merged) one follow.

* Enabling antialiasing results in the vertical separator lines being blurred into two pixels instead of being exactly 1-pixel wide. This is not good. (I'd actually prefer the arrows to also be 1-pixel wide, but that's up to you and @eteran to decide.)
* If a jump targets into the middle of an instruction, the arrow line is drawn to the bottom of the viewport, missing the line containing the target byte.

The following screenshot demonstrates both above mentioned problems:
Screenshot - 251019 - 00:03:39

* The arrows in the following screenshot look confusing:
Screenshot - 251019 - 00:14:03

* The arrows for untaken branches are asymmetric with respect to their path line:
Screenshot - 251019 - 00:17:10

* The following display is misleading: I've actually come to this state by stepping from 0xf7fd7817 (skipping the conditional jump), but the red arrow hints as if I have just jumped here from 0xf7fd77fb. Maybe the arrow from the origin shouldn't be painted red.
Screenshot - 251019 - 00:20:45

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
+						j == jump_arrow_prev.horizontal_width) {
+						is_width_good = false;
+						break;
+					}
+				}
 
-				arrow_horizontal_length = width_downward_outviewport + font_width_ * 2;
-				width_downward_outviewport = arrow_horizontal_length; // update new width
+				// `j` horizontal width is not good, search next
+				if (is_width_good == false) {
```
10110111: Why not `!is_width_good`?

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
 			return a.distance < b.distance;
 		});
 
-		int width_upward_outviewport = 0;
-		int width_downward_outviewport = 0;
+		// find suitable arrow horizontal width
+		for (int i = 0; i < jump_arrow_vec.size(); i++) {
```
10110111: I'd prefer the `i`, `j` and `k` variables to be named more descriptively. It's hard to keep in mind what three one-letter variables represent when trying to follow the logic.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
-				arrow_horizontal_length = jump_arrow.distance * (font_width_ / 3);
+				bool is_width_good = true;
 
-			} else if (is_dst_upward) {
+				// check if width clash with previous jump arrow
+				for (int k = 0; k < i; k++) {
 
-				arrow_horizontal_length = width_upward_outviewport + font_width_ * 2;
-				width_upward_outviewport = arrow_horizontal_length;  // update new width
+					JumpArrow& jump_arrow_prev = jump_arrow_vec[k];
```
10110111: This reference should be to `const`: you aren't going to change `jump_arrow_vec[k]`.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
+					bool is_dst_upward_prev = jump_arrow_prev.target < instructions_[jump_arrow_prev.src_line].rva();
 
-			} else {
+					int jump1_min = std::min(
+						jump_arrow.src_line, 
+						jump_arrow.dst_in_viewport ? jump_arrow.dst_line : (is_dst_upward ? 0 : viewport()->height())
```
10110111: Why not make a variable for this long subexpression, which is also repeated in the next `max` call? Same for the `jump2_min` and `jump2_max` min/max call arguments.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
+						jump_arrow_prev.src_line, 
+						jump_arrow_prev.dst_in_viewport ? jump_arrow_prev.dst_line : (is_dst_upward_prev ? 0 : viewport()->height())
+					);
+					
+					int jump2_max = std::max(
+						jump_arrow_prev.src_line,
+						jump_arrow_prev.dst_in_viewport ? jump_arrow_prev.dst_line : (is_dst_upward_prev ? 0 : viewport()->height())
+					);
+
+					bool cond1 = jump2_max >= jump1_max && jump2_min >= jump1_max;
```
10110111: If I understand this logic correctly, you are checking whether whole jump2 arrow is below jump1 one. This is not obvious until the reader analyzes the expression. So the name of `cond1` should be more descriptive. Same concern for `cond2` in the next line.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
+					bool cond2 = jump2_min <= jump1_min && jump2_max <= jump1_min;
+					bool is_jump_overlap = !(cond1 || cond2);
```
10110111: Maybe `jumps_overlap` instead of `is_jump_overlap`. It would sound a bit more natural.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1285,38 +1286,85 @@ void QDisassemblyView::paintEvent(QPaintEvent *) {
...
+						jump_arrow_prev.src_line,
+						jump_arrow_prev.dst_in_viewport ? jump_arrow_prev.dst_line : (is_dst_upward_prev ? 0 : viewport()->height())
+					);
+
+					bool cond1 = jump2_max >= jump1_max && jump2_min >= jump1_max;
+					bool cond2 = jump2_min <= jump1_min && jump2_max <= jump1_min;
+					bool is_jump_overlap = !(cond1 || cond2);	
+
+					if (is_jump_overlap && 
+						j == jump_arrow_prev.horizontal_width) {
```
10110111: Maybe avoid splitting this line? It seems short enough when merged (especially compared to the ternary conditionals above).

src/widgets/QDisassemblyView.cpp
shahril96: I thought using  "== false" looks more readable in this case. But then, I will fix it. :)

src/widgets/QDisassemblyView.cpp
shahril96: I've renamed all these three variables.

src/widgets/QDisassemblyView.cpp
shahril96: Yes, you're correct, I've fixed that.

src/widgets/QDisassemblyView.cpp
shahril96: Fixed.

src/widgets/QDisassemblyView.cpp
shahril96: I've added comments to clarify what the comparison is doing.

src/widgets/QDisassemblyView.cpp
shahril96: Fixed.

src/widgets/QDisassemblyView.cpp
shahril96: Fixed.

shahril96:
1)

image

Arrow there is red highlighted because there is a direct jump from 0xf7fd77fd, the reason why I do highlight it is to give a clarification that this jump to this address is always taken.

2)

image

I've fixed this by adding more horizontal length to one of the jump blocks if it is overlap at the edge.

3)

image

Fixed where jump now points to the middle of instruction (see screenshot above)

4)

I have removed anti-aliasing code.
If @eteran has other to say I will add it back.

Thank you for all the comments, really appreciated it. :)

10110111:
> I have removed anti-aliasing code.

I think the arrow tips could benefit from antialiasing. It just has to be enabled in `QPainter` right before that code and disabled afterwards. Just be sure to make them still look symmetric (note that in antialiased case the samples are offset by 0.5 px).

10110111:
I'm OK with these changes now. I was thinking of some pixel-hunting tweaks, but I suppose I'll simply make a follow-up PR when I finalize my ideas.

eteran:
@10110111 's review has proven to be very worth having! Thanks to both of you. This PR looks great to me. I'll probably merge later today.

PR #731, opened 2019-11-01
Add option to disable jump draw

shahril96:
Changes made:

1. Added option to disable jump draw

image

2. Fixed possible crashing when accessing show_addresses_ function 

3. Dock registers badge at the left side of line1

image

Any comments is really appreciated. :)

### DIFF
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -1283,7 +1283,7 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 		return a.distance < b.distance;
 	});
 
-	auto isLineOverlap = [&](int line1_head, int line1_tail, int line2_head, int line2_tail, bool edge_overlap = true) -> bool {
+	auto isLineOverlap = [&](int line1_head, int line1_tail, int line2_head, int line2_tail, bool edge_overlap) -> bool {
 
 		int jump1_arrow_min = std::min(line1_head, line1_tail);
 		int jump1_arrow_max = std::max(line1_head, line1_tail);
@@ -1725,7 +1725,7 @@ int QDisassemblyView::line1() const {
 
 		// allocate space for register badge
 		// 4 (maximum register name for GPR) + overhead
-		return (edb::v1::config().show_register_badges) ? (5 * font_width_ + font_width_/2) : 0;
+		return edb::v1::config().show_register_badges ? (5 * font_width_ + font_width_/2) : 0;
 
 	} else if(line1_ == 0) {
 		return 15 * font_width_;
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -1370,7 +1370,8 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 		// if direct jmp is selected, then draw arrow in red
 		if (is_unconditional_jump(instructions_[jump_arrow.src_line]) && 
 			(ctx->selected_line == jump_arrow.src_line || 
-			(ctx->selected_line == jump_arrow.dst_line && show_addresses_[jump_arrow.dst_line] != current_address_ ))) {
+			(ctx->selected_line == jump_arrow.dst_line && 
+			(jump_arrow.dst_in_viewport && show_addresses_[jump_arrow.dst_line] != current_address_ )))) {
 			arrow_color = Qt::red;
 		}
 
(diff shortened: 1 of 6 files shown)


src/widgets/QDisassemblyView.cpp
```diff
@@ -1370,7 +1370,8 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 		// if direct jmp is selected, then draw arrow in red
 		if (is_unconditional_jump(instructions_[jump_arrow.src_line]) && 
 			(ctx->selected_line == jump_arrow.src_line || 
-			(ctx->selected_line == jump_arrow.dst_line && show_addresses_[jump_arrow.dst_line] != current_address_ ))) {
+			(ctx->selected_line == jump_arrow.dst_line && 
+			(jump_arrow.dst_in_viewport && show_addresses_[jump_arrow.dst_line] != current_address_ )))) {
```
10110111: Nitpick: the additional parentheses are superfluous here, and they do add to clutter on the closing side: `))))`. I'd suggest removing this pair, retaining the logically related conditions on the single line as they are now.

10110111:
When I set _Show jump arrows_ and unset _Show when registers are pointing to a visible address_, the space allocated appears to be large enough for both, although only one thing is enabled.

src/widgets/QDisassemblyView.cpp
shahril96: Fixed

shahril96:
@10110111, that is an intended behavior, as register badge is drawn inside jump arrow area (at left side of first line)

10110111:
It's not drawn when it's disabled, is it?

shahril96:
Peek 2019-11-02 03-48

Register badge is drawn inside jump arrow area, so there is no effect on area allocated if either the badge is drawn or not.

Jump arrow area is always going to be initially as this size (https://github.com/eteran/edb-debugger/pull/731/files#diff-4af38e37cb69a85d76c168f3230a7adeL1650)

Unless I got you wrong 😅

10110111:
Ah, so you made it more economical... But don't you think it's not very nice when the arrows go over the badges, like in the following screenshot?
Screenshot - 011119 - 23:09:30

shahril96:
@10110111 that is a good catch (doesn't appear in my test 😅)

thank you! will fix it soon

src/widgets/QDisassemblyView.cpp
```diff
@@ -1647,11 +1676,19 @@ int QDisassemblyView::line0() const {
 // Desc:
 //------------------------------------------------------------------------------
 int QDisassemblyView::line1() const {
-	if(line1_ == 0) {
+
+	if(!edb::v1::config().show_jump_arrow) {
+
+		// allocate space for register badge
+		// 5 (maximum register name length i can find off) + overhead
```
10110111: Which of the registers has 5-char name? `MXCSR`? `YMM12`? I would be surprised if they were able to point anywhere ;)
Only GPRs are used for the badges, see the beginning of `drawRegiserBadges()`.

shahril96: Fixed

shahril96:
@10110111 latest commits fixed issues that you pointed out

src/widgets/QDisassemblyView.cpp
```diff
@@ -1283,12 +1283,35 @@ void QDisassemblyView::drawJumpArrows(QPainter &painter, const DrawingContext *c
 		return a.distance < b.distance;
 	});
 
+	auto isLineOverlap = [&](int line1_head, int line1_tail, int line2_head, int line2_tail, bool edge_overlap = true) -> bool {
```
10110111: You are not using this default `= true`, so no point in it.

src/widgets/QDisassemblyView.cpp
```diff
@@ -1689,8 +1724,8 @@ int QDisassemblyView::line1() const {
 	if(!edb::v1::config().show_jump_arrow) {
 
 		// allocate space for register badge
-		// 5 (maximum register name length i can find off) + overhead
-		return (edb::v1::config().show_register_badges) ? (6 * font_width_ + font_width_/2) : 0;
+		// 4 (maximum register name for GPR) + overhead
+		return (edb::v1::config().show_register_badges) ? (5 * font_width_ + font_width_/2) : 0;
```
10110111: Why parentheses around single variable?

10110111:
In other respects, this looks fine to me now.

shahril96:
Pushed fixes as mentioned by @10110111

src/widgets/QDisassemblyView.cpp
shahril96: fixes

src/widgets/QDisassemblyView.cpp
shahril96: fixed

eteran:
Looks good to me! Merging!

@10110111 Thanks for the code review, @shahril96 thanks for this awesome contribution!

PR #733, opened 2019-11-09
Style fixup

eteran:
A ton of commits to clean up the style (mostly) and address a few bugs.
It also fixes ARM support and Windows builds, though I'm sure it could use a lot more work.
Fixes assumptions made about capstone version (we were using symbols only available in version 4.x)
Working on getting things to a point where the auto-formatter can be used (`.clang-format` included), it's pretty close and has been applied to the codebase, but I'm sure that we'll run into a few spots where it made some questionable formatting decisions... We'll fix em as we find em.
Started migration to Doxygen comment block style. I like my old style aesthetically a lot better... but the convenience of my editor being able to automatically add them is just too much to pass up.


eteran:
@10110111 Just a heads up on this one. LOTS of changes, and it's certainly not perfect from a formatting point of view, but it's a lot of progress as far as consistent style, formatting, etc...

I plan to do a bunch more work towards this but I just didn't want this PR that touches 400 files to "come out of nowhere", lol

plugins/Analyzer/AnalyzerWidget.cpp
```diff
@@ -17,29 +17,32 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+#include <QScrollBar>
 
 namespace AnalyzerPlugin {
 
-//------------------------------------------------------------------------------
-// Name:
-//------------------------------------------------------------------------------
-AnalyzerWidget::AnalyzerWidget(QWidget *parent, Qt::WindowFlags f) : QWidget(parent, f)  {
+/**
+ * @brief AnalyzerWidget::AnalyzerWidget
```
10110111: Don't you find that such placeholders that contain no substance, just repeating the obvious, are useless? I think it'd be better to create them whenever you're really going to make useful comments (document the function), otherwise these are just like dust in the source code — useless but unavoidable (besides, Style Guide doesn't appear to require them, so they seem actually avoidable).

plugins/InstructionInspector/Plugin.cpp
```diff
@@ -1511,7 +1511,7 @@ InstructionDialog::InstructionDialog(QWidget *parent, Qt::WindowFlags f)
 			// TODO: mem_barrier
 			add({"Operands"});
 
-			auto *const operands = tree->topLevelItem(tree->topLevelItemCount() - 1);
+            auto *const operands = tree_->topLevelItem(tree_->topLevelItemCount() - 1);
```
10110111: I don't quite get it: are we switching from tabs to spaces? BTW, the Style Guide has no mention of them.

plugins/ODbgRegisterView/DialogEditFPU.cpp
```diff
@@ -38,7 +38,7 @@ long double readFloat(const QString &strInput, bool &ok) {
 	ok = false;
 	const QString str(strInput.toLower().trimmed());
 
-	if(const auto value = util::fullStringToFloat<long double>(str.toStdString())) {
+	if (const auto value = util::fullStringToFloat<long double>(str.toStdString())) {
```
10110111: I think this should be put into the Style Guide. Currently not only doesn't it tell to put space after `if`, but its examples also don't follow this rule.

plugins/ODbgRegisterView/DialogEditFPU.cpp
```diff
@@ -76,7 +76,8 @@ long double readFloat(const QString &strInput, bool &ok) {
 }
 }
 
-DialogEditFPU::DialogEditFPU(QWidget *parent, Qt::WindowFlags f) : QDialog(parent, f), floatEntry_(new ODbgRegisterView::Float80Edit(this)), hexEntry_(new QLineEdit(this)) {
+DialogEditFPU::DialogEditFPU(QWidget *parent, Qt::WindowFlags f)
+	: QDialog(parent, f), floatEntry_(new ODbgRegisterView::Float80Edit(this)), hexEntry_(new QLineEdit(this)) {
```
10110111: Maybe add a rule to format initialization lists in a column? I suggest, e.g. something like this:
```
	: QDialog(parent, f)
    , floatEntry_(new ODbgRegisterView::Float80Edit(this))
    , hexEntry_(new QLineEdit(this))
{
```
(Yes, with a deviation from the general rule about K&R braces, because this brace really gets lost in the parentheses otherwise.)

10110111:
Tabs vs spaces is a mess after some of these changes... But there appears to still be an overall improvement: 37651 instances of `^\t` and 350 instances of `^ {4}` in `style-fixup`, compared to 35587 and 1515 respectively in current `master`:
```
$ git status | head -1
HEAD detached at upstream/style-fixup
$ find . \( -name '*.cpp' -o -name '*.h' -o -name '*.hpp' -o -name '*.cc' \
              -o -name '*.cxx' -o -name '*.hxx' \) -exec grep -P '^\t' {} + | wc -l
37651
$ find . \( -name '*.cpp' -o -name '*.h' -o -name '*.hpp' -o -name '*.cc' \
               -o -name '*.cxx' -o -name '*.hxx' \) -exec grep -P '^ {4}' {} + | wc -l
350
```

plugins/InstructionInspector/Plugin.cpp
eteran: That would be an auto-format error. Sometimes clang-format gets confused and things it's aligning with spaes in certain contexts. fix this one up

eteran: Actually, it is tabs in the latest update to the PR.

plugins/ODbgRegisterView/DialogEditFPU.cpp
eteran: Yes, the style guide will be updated to reflect how the code is actually formatted, good call. Usually, I address the parens getting lost by putting a newline after the initializer list, but I'm willing to take this under consideration as a better solution.

plugins/ODbgRegisterView/DialogEditFPU.cpp
eteran: Agreed, it is however enforced by clang-format :+1: . It will be added to the style guide.

plugins/Analyzer/AnalyzerWidget.cpp
eteran: So, at least for me, they serve 2 primary purposes.

1. While not immediately having substance, they do provide a designated place to put any information we need to know about the function, should we need it. It's a consistent place to put it.

2. but more importantly, I really like having a visual break between functions that is unmistakable, it helps a lot with navigation when scrolling to have a different colored comment block dividing up functions so they don't blend together. I loved the style I learned in college with the wall of dashes, but my editor does Doxygen comments in like an instant, so it's hard to argue with that level of convinience.

eteran:
**note** almost all of the leading spaces are in GPL comment blocks! There is a handful worth fixing up though.

eteran:
```
find . \( -name '*.cpp' -o -name '*.h' -o -name '*.hpp' -o -name '*.cc' -o -name '*.cxx' -o -name '*.hxx' \) -exec grep -P '^ {4}' {} +  | grep -v "@" | grep -v "Free Software"
```

Nice and clean with tabs vs spaces now

eteran:
@10110111 Unless you have an objection, I plan to merge sometime this afternoon (EDT). I'll keep an eye out for any more comments from you though.

PR #749, opened 2020-02-18
fix warning message printed in ODBRegView::modelReset

sorokin:
After commit 562eb3c2bd3853ac53e031963dd38a744367f821 a message

```
   QLayout: Attempting to add QLayout "" to ODbgRegisterView::ODBRegView "ODBRegView", which already has a layout
```

was printed everytime ODBRegView::modelReset is called.

This commit tries to fix this.


eteran:
Seems reasonable to me.

Thanks!

sorokin:
Thank you!

Issue #752, opened 2020-02-22
address and instruction bytes column overlap after reopening


sorokin:
If I make address column too small and then close edb and then open it again, I get disassembly view that looks like this:

![disassembly-view-after-load](https://user-images.githubusercontent.com/286877/75096889-24057380-55b5-11ea-994e-cf23f31a5541.png)

If I trace the values of l1, l2, l3, l4 in DrawingContext I will get something like this: (165 380 231 1194), while before reopening it was (165 190 223 11940). As I found out the line2_ loaded correctly in QDisassemblyView::restoreState and then is reset to 0 in QDisassemblyView::setRegion.

At first my diff to fix the problem was:
```
$ git diff --cached
diff --git a/src/widgets/QDisassemblyView.cpp b/src/widgets/QDisassemblyView.cpp
index fc44ae2d..eaafca09 100644
--- a/src/widgets/QDisassemblyView.cpp
+++ b/src/widgets/QDisassemblyView.cpp
@@ -535,10 +535,6 @@ void QDisassemblyView::setRegion(const std::shared_ptr<IRegion> &r) {
                setAddressOffset(region_ ? region_->start() : edb::address_t(0));
                updateScrollbars();
                Q_EMIT regionChanged();
-
-               if (line2_ != 0 && line2_ < autoLine2()) {
-                       line2_ = 0;
-               }
        }
        update();
 }
```

You might compare it with this commit 8f2c83137a4432a948778c8d9b5c6bf6ed905e57. :-)

Now the question is what is the proper way of fixing the problem. One might argue that writing line2_ without updating line3_ and line4_ breaks class invariants. And updating them too would fix the visual glitch, but I would say that resetting column widths set by user is not a good behavior.

To fix the visual glitch, but also to address #492, I would recomment having different column width settings for 32 and for 64 bit modes.

PR #753, opened 2020-02-23
remove variable that is always equal to 0

sorokin:


eteran:
I'm not sure that it's a benefit to remove this particular variable. While there is nothing to the left of "line 0", there could one day be, so having it be a tunable parameter seems to at the very least, do no harm.

sorokin:
Let me explain why I decided to remove this variable.

I wanted to fix #752 by making separate settings for line{1,2,3,4} for 32- and 64-bit modes. Apparently line{1,2,3,4} are saved to settings while line0 is not. So I decided to remove this asymmetry are removed line0.

But you are right for now line0 can stay.

eteran:
Gotcha makes sense about the settings. Thanks for your efforts in fixing these bugs.

I'm going to close this PR, if it turns out that in the future, there is a tangible benefit to removing this variable, feel free to open another :-).

PR #761, opened 2020-05-03
Merge pull request #1 from eteran/master

gxllang83:
update


10110111:
This seems to be doing nothing to the EDB master, only creating this merge commit.

ped7g:
I have suspicion it may be accidental pull request by somebody using git rarely, trying to update his fork with this parent, ending with merge commit, and then trying to "sync" it to remove any remaining differences? I can imagine situation like that, git is not completely trivial to understand and to pick up best-practice flows. If it is not like this, then I'm completely missing the point of this PR.

eteran:
Seems to be in error. Closing.

PR #763, opened 2020-05-19
WIP: Session manager update

Vic063:
These commits enable a better support of the Session Manager by now saving:
- labels
- breakpoints

This is a WIP patch as I don't know if my code/modifications fit your needs; do not hesitate to tell me what's wrong and how I can improve this patch.

On a side note, I wanted to know if it won't be better, instead of saving the fixed address of the comment/label/breakpoint, to save the relative address of the comment/label/breakpoint with the name of the module so the restore process should correctly work even if base addresses change at each process restart (with ASLR for example). What do you think about this idea?

Thanks.

### DIFF
--- a/include/ISymbolManager.h
+++ b/include/ISymbolManager.h
@@ -42,6 +42,7 @@ public:
 	virtual void loadSymbolFile(const QString &filename, edb::address_t base)          = 0;
 	virtual void setSymbolGenerator(ISymbolGenerator *generator)                       = 0;
 	virtual void setLabel(edb::address_t address, const QString &label)                = 0;
+    virtual void restoreLabels(const QVariantList &labels)                             = 0;
 	virtual QString findAddressName(edb::address_t address, bool prefixed = true)      = 0;
 	virtual QHash<edb::address_t, QString> labels() const                              = 0;
 	virtual QStringList files() const                                                  = 0;
--- a/include/ISymbolManager.h
+++ b/include/ISymbolManager.h
@@ -42,7 +42,7 @@ public:
 	virtual void loadSymbolFile(const QString &filename, edb::address_t base)          = 0;
 	virtual void setSymbolGenerator(ISymbolGenerator *generator)                       = 0;
 	virtual void setLabel(edb::address_t address, const QString &label)                = 0;
-    virtual void restoreLabels(const QVariantList &labels)                             = 0;
+	virtual void restoreLabels(const QVariantList &labels)                             = 0;
 	virtual QString findAddressName(edb::address_t address, bool prefixed = true)      = 0;
 	virtual QHash<edb::address_t, QString> labels() const                              = 0;
 	virtual QStringList files() const                                                  = 0;
--- a/src/Debugger.cpp
+++ b/src/Debugger.cpp
@@ -1013,7 +1013,7 @@ void Debugger::closeEvent(QCloseEvent *event) {
 	// make sure sessions still get recorded even if they just close us
 	const QString filename = sessionFilename();
 	if (!filename.isEmpty()) {
-		SessionManager::instance().saveSession(filename);
+        edb::v1::session_manager().saveSession(filename);
 	}
 
 	if (IDebugger *core = edb::v1::debugger_core) {
@@ -2829,7 +2829,7 @@ void Debugger::detachFromProcess(DetachAction kill) {
 
 	const QString filename = sessionFilename();
 	if (!filename.isEmpty()) {
-		SessionManager::instance().saveSession(filename);
+        edb::v1::session_manager().saveSession(filename);
 	}
 
 	programExecutable_.clear();
@@ -2889,11 +2889,13 @@ void Debugger::setInitialDebuggerState() {
 	const QString filename = sessionFilename();
 	if (!filename.isEmpty()) {
 
-		SessionManager &session_manager = SessionManager::instance();
+        const ISessionManager &session_manager = edb::v1::session_manager();
 
-		if (Result<void, SessionError> session_error = session_manager.loadSession(filename)) {
-			QVariantList comments_data = session_manager.comments();
-			ui.cpuView->restoreComments(comments_data);
+        if (Result<void, SessionError> session_error = edb::v1::session_manager().loadSession(filename)) {
+            QVariantList comments_data = session_manager.comments();
+            QVariantList labels_data = session_manager.labels();
+            ui.cpuView->restoreComments(comments_data);
+            edb::v1::symbol_manager().restoreLabels(labels_data);
 		} else {
 			QMessageBox::warning(
 				this,
--- a/src/Debugger.cpp
+++ b/src/Debugger.cpp
@@ -2888,14 +2888,24 @@ void Debugger::setInitialDebuggerState() {
 
 	const QString filename = sessionFilename();
 	if (!filename.isEmpty()) {
+		std::shared_ptr<IBreakpoint> bp;
+		const ISessionManager &session_manager = edb::v1::session_manager();
 
-        const ISessionManager &session_manager = edb::v1::session_manager();
+		if (Result<void, SessionError> session_error = edb::v1::session_manager().loadSession(filename)) {
+			QVariantList breakpoints_data = session_manager.breakpoints();
+			QVariantList comments_data = session_manager.comments();
+			QVariantList labels_data = session_manager.labels();
+			ui.cpuView->restoreComments(comments_data);
+			edb::v1::symbol_manager().restoreLabels(labels_data);
+			for (auto it = breakpoints_data.begin(); it != breakpoints_data.end(); ++it) {
+				IBreakpoint::TypeId type;
+				QVariantMap bp_data = it->toMap();
 
-        if (Result<void, SessionError> session_error = edb::v1::session_manager().loadSession(filename)) {
-            QVariantList comments_data = session_manager.comments();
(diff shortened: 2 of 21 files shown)


src/widgets/QDisassemblyView.h
```diff
@@ -82,9 +82,9 @@ class QDisassemblyView final : public QAbstractScrollArea {
 	int removeComment(edb::address_t address);
 	int selectedSize() const;
 	std::shared_ptr<IRegion> region() const;
-	void addComment(edb::address_t address, QString comment);
-	void clearComments();
-	void restoreComments(QVariantList &);
+    void addComment(edb::address_t address, QString comment);
```
10110111, line 85: Nothing seems to have changed here except whitespace?..

src/Debugger.cpp
```diff
@@ -2888,14 +2888,24 @@ void Debugger::setInitialDebuggerState() {
...
             ui.cpuView->restoreComments(comments_data);
             edb::v1::symbol_manager().restoreLabels(labels_data);
+            for (auto it = breakpoints_data.begin(); it != breakpoints_data.end(); ++it) {
+                IBreakpoint::TypeId type;
+                QVariantMap bp_data = it->toMap();
+
+                type = static_cast<IBreakpoint::TypeId>(bp_data["type"].toInt());
+                bp = edb::v1::create_breakpoint(edb::v1::string_to_address(bp_data["address"].toString()).value());
+                bp_data["enabled"].toBool() ? bp->enable() : bp->disable();
+                //bp->setType(type);
```
10110111, line 2907: Is there any problem with this call?

src/session/SessionManager.cpp
```diff
@@ -247,23 +282,43 @@ void SessionManager::addLabel(const Label &l) {
...
 * @param address
 */
-void SessionManager::removeComment(edb::address_t address) {
+void SessionManager::removeBreakpoint(edb::address_t address) {
 	QString hexAddressString   = address.toHexString();
-	QVariantList comments_data = sessionData_["comments"].toList();
+    QVariantList breakpoints_data = sessionData_["breakpoints"].toList();
 
-	auto it = std::find_if(comments_data.begin(), comments_data.end(), [&hexAddressString](QVariant entry) {
+    auto it = std::find_if(breakpoints_data.begin(), breakpoints_data.end(), [&hexAddressString](QVariant entry) {
```
10110111, line 292: Can't we have multiple breakpoints at the same address? E.g. with different conditions, different kinds of access... (Maybe not, I don't know.)

include/ISymbolManager.h
```diff
@@ -42,6 +42,7 @@ class ISymbolManager {
 	virtual void loadSymbolFile(const QString &filename, edb::address_t base)          = 0;
 	virtual void setSymbolGenerator(ISymbolGenerator *generator)                       = 0;
 	virtual void setLabel(edb::address_t address, const QString &label)                = 0;
+    virtual void restoreLabels(const QVariantList &labels)                             = 0;
```
10110111, line 45: Here and in many other places: please note that EDB uses 4-char tabs, not spaces, for indentation.

src/edb.cpp
```diff
@@ -202,6 +203,15 @@ ISymbolManager &symbol_manager() {
 	return g_SymbolManager;
 }
 
+//------------------------------------------------------------------------------
+// Name: session_manager
+// Desc:
+//------------------------------------------------------------------------------
+ISessionManager &session_manager() {
+	static SessionManager g_SessionManager = SessionManager::instance();
```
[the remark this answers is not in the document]
Vic063: To be honest, I mainly reproduced what other methods which look like this one do.
But you are right, I will try as you said and come back to resolve this.

src/widgets/QDisassemblyView.h
Vic063: Exactly, I think it's a mistake made by copying/pasting this line to create the 2 next methods and the different tools I used to write the whole patch (started with nano for Python profile and switched back on QtCreator to enjoy the power of the Qt engine before trying to compile).

src/Debugger.cpp
Vic063: When trying this patch, the call to the setType method completely hangs EDB for an unknown reason. Il will look forward into this issue and bring a solution.

src/session/SessionManager.cpp
Vic063: I agree with you. I think as I did this quickly, I followed the same logic used by comments/label.

include/ISymbolManager.h
Vic063: OK! Sorry for that, I opened the EDB source code today, for the first time. I started this patch using nano with a specific profile for Python and switched to QtCreator.

I will correct that during the next commits.

Thank you for your quick review!

src/edb.cpp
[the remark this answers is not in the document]
Vic063: Resolved in the force-pushed update.

src/widgets/QDisassemblyView.h
Vic063: Resolved with the last commit.

include/ISymbolManager.h
Vic063: Resolved with the last commit.

src/Debugger.cpp
Vic063: Fixed in the last force-pushed update.
Note : calling SetType with a IBreakpoint::TypeId causes an infinite loop.

eteran:
A lot of activity on this PR! Awesome, let us know when you think it's ready for review :-)

Vic063:
Ok, I was thinking about a new strategy to handle a session loading/saving that I started to implement before going to a place without any technology (holidays).

The big 'problem' I encounter right now is how to properly save the breakpoint state, which would require to rewrite a lot of stuff I think.

eteran:
@Vic063 can you elaborate a bit more on the difficulties you have with the current design?

Any plugin should be able to save and load more or less arbitrary data. It just needs to be serialized into something that the session manager can use.

Vic063:
@eteran here is my approach to handle session 'objects' (labels, comments, breakpoints, etc...): https://github.com/Vic063/edb-debugger/commit/691532d7077596a92b14cdaa0d14da8c3e062003

A session object can have different kind of type but for the session manager, it is always treated as a 'SessionObject'. An object is always linked to a module, has a relative address and custom properties according to its type.
The linked commit shows an example of my approach with labels and comments (I did not have time to implement breakpoint type).

Here is my attempt to write it down on a paper, using my poor UML knowledge:


When the session file is saved, it looks like this:
`{
    "id": "edb-session",
    "objects": {
        "0000000000001090": {
            "comment": "Libc entry point",
            "module": "/lib/x86_64-linux-gnu/ld-2.28.so",
            "type": "comment"
        },
        "00000000000010c5": {
            "comment": "Test Libc2",
            "module": "/lib/x86_64-linux-gnu/ld-2.28.so",
            "type": "comment"
        },
        "0000000000001e90": {
            "label": "Label1_Test",
            "module": "/lib/x86_64-linux-gnu/ld-2.28.so",
            "type": "label"
        },
        "000000000000f400": {
            "label": "Label2_Test",
            "module": "/lib/x86_64-linux-gnu/ld-2.28.so",
            "type": "label"
        }
    },
    "plugin-data": {
        "BookmarksPlugin::Bookmarks": {
            "bookmarks": [
            ]
        }
    },
    "timestamp": "2020-05-28T20:24:49.880Z",
    "version": 1
}`

And in the rendering in EDB:
screenshot_edb

I don't know if it's clear to you?

eteran:
@Vic063 Sorry for the delay in response. I suppose my question is, can't you implement that structure using the current API? Can't you just have the session manager or a given plugin save an "objects" key and have whatever dictionary you like under it?

Am I missing something?

eteran:
That being said, I don't object to this, or any other sensible data format we come up with :-)

Vic063:
@eteran In my mind, I thought to a mechanism where the Session manager does not have to know which object it has to save/restore, it is completely transparent to it.

On the other side, the intelligence resides in the object itself, which has its own saving/restore mechanism.

Using this method, when declaring a new object that needs to be saved, it would require only 2/3 lines/methods to add to the session manager and the whole logic would be written in a separate class file. I think it might decrease the risk of breaking the saving format and simplify the task when someone want to add or modify a session object.
Moreover, one of the issue I encountered (if I remember correctly) when I wanted to properly implement breakpoints saving, is the 2-steps saving update (here is what is currently done) : 

1. First, we create the breakpoint to the given address, it is saved in the session manager
2. We want to update the breakpoint to add a break condition, it is not saved, and to save it, we have to find the breakpoint in the session manager remove it and save it again with its new property (the break condition)

With what I purpose, the breakpoint is a specific object that can listen to some signals, like the 'update' one, which can be triggered when adding/removing a condition or something else. This would not require to manipulate the session manager until the debugger is closed.

My approach might be a little bit tricky to understand, but I think this is due to the fact that I don't handle comments/labels/breakpoints like specific plugins, but like objects that are strongly attached to the debugger. In my mind, a plugin is something we can use or not according to our usage and I can't imagine using a debugger without comments/labels/breakpoints :)

eteran:
@Vic063 I'm closing this PR due to lack of activity. But take heart, We are currently discussing moving breakpoint management out of plugins and into the core binary because it is so fundamental as you point out.

Please re-open, or open a new PR if you want to continue efforts on this.

PR #771, opened 2020-08-18
Don't save default plugin path

maxxk:
In Linux distributions like NixOS and Guix installation path contains hash of all dependencies.
If plugin path is stored in user configuration, edb will break on update due to invalid plugin path.
It can be prevented by saving only non-default plugin path.

Issue is described here: https://github.com/NixOS/nixpkgs/pull/58962#issuecomment-481521588

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -274,6 +274,8 @@ void Configuration::writeSettings() {
 	settings.setValue("directory.symbol.path", symbol_path);
 	if (plugin_path != getDefaultPluginPath()) {
 		settings.setValue("directory.plugin.path", plugin_path);
+	} else {
+		settings.remove("directory.plugin.path");
 	}
 	settings.setValue("directory.session.path", session_path);
 	settings.endGroup();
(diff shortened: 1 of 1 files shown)


eteran:
Generally looks good to me. So it won't save it so long as it's the default, but if it's user set, then it will still save it, right?

maxxk:
Well, it will save it if it is different from the default. When rechecking it just now I found an edge case: if plugin path in config file is incorrect, then my changes make it impossible to replace it with the default path. I can't think of a simple enough way to check whether user entered the default value. Probably I will add an "else" clause to `writeSettings` which will remove plugin path from settings if the default path is entered. It will remove an invalid path and after the restart plugins will be loaded from the default path.

eteran:
OK, I'll hold off on merging if you want to try to address any edge cases you've found.

maxxk:
Thank you, I pushed the fix.

eteran:
@10110111 This patch looks good to me, you have any thoughts?

10110111:
Looks sane to me.

PR #790, opened 2020-12-17
Make it possible to easily restore a hidden register group

### LINKED ISSUE #708: No way to show CPU registers after hiding them
If you hide Registers there's no way to get them back. I even tried to hide them all to see what happened and now I'm with a blank window where the registers should be.

10110111:
Fixes #708.


eteran:
Simple and works. Looks good to me!

PR #792, opened 2020-12-21
Adds support for using a builtin gdtoa-desktop if not found as a system install

eteran:
First tries a system-level install, if not found, will build the internal copy of the library (won't build if installed on the system already and was found).

Because gdtoa-desktop is based on some old code that we are striving to keep clean of any unnecessary changes, it would normally build with lots of warnings if we applied the same flags we doe for edb to it, so also improved our ability to be selective with which parts we apply which flags to.


eteran:
@10110111 I think this is working correctly!

Can you confirm that on your system it does **not** build gdtoa-desktop, and instead just uses the system-installed copy?

10110111:
Links to the system library on my system:
```
$ ldd `which edb` | grep gdtoa
        libgdtoa-desktop.so => /usr/lib/libgdtoa-desktop.so (0xf7dcb000)
```

eteran:
Aiight, I'm making it so windows builds won't try to use the library... yet. Assuming that passes, I'm planning on merging this in. Next on the docket would be capstone so we can say "if you have >= 4.0 on your system we'll use it; if you have < 4.0, or you no capstone at all, we'll use the built-in one"

PR #796, opened 2020-12-21
Windows TravisCI builds...

eteran:
Would be cool if we could get windows to work on travis too!


eteran:
No need anymore, github CI easily supports windows builds, we'll just use that exclusively over travis.

PR #814, opened 2021-12-12
Add Fedora RPM specfile

brndd:
This is a specfile for packaging edb into an RPM for Fedora. Other distros (eg. OpenSUSE) are untested; it certainly won't build on OpenSUSE due to missing dependencies.

I've used this to package edb on COPR (the Fedora user repository) for personal use, but figured I might as well put it here too in case it interests you. With this it should be relatively simple to eg. set up a COPR that automatically builds every GitHub release as soon as you push them, using COPR's webhook functionality.

Building the RPM locally using this is relatively simple:

1. Run `rpmdev-setuptree` to set up the rpmbuild directory in your ~/.
2. Move edb.spec to `~/rpmbuild/SPECS/`
3. Build the RPM with `cd ~/rpmbuild && rpmbuild -ba SPECS/edb.spec`

### DIFF
--- a/pkg/fedora/edb.spec
+++ b/pkg/fedora/edb.spec
@@ -1,0 +1,206 @@
+Name:           edb
+Version:        1.3.0
+Release:        2%{?dist}
+Summary:        A debugger based on the ptrace API and Qt
+
+License:        GPLv2+
+URL:            https://github.com/eteran/edb-debugger
+Source0:        https://github.com/eteran/edb-debugger/releases/download/%{version}/edb-debugger-%{version}.tgz
+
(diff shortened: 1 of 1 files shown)


eteran:
Thanks!

brndd:
Oh, I should mention: currently this only builds on x86_64 because of this line: `ExclusiveArch:  %{ix86} x86_64`. AArch32 and 32-bit x86 aren't supported by Fedora, but if edb ever gets AArch64 support then this specfile should be updated to match that.

eteran:
Good to know, technically edb DOES already have AArch64 support, but it probably could use a bit of improvement

PR #820, opened 2022-06-20
make binary string search read proc mem in chunks

AaronOpfer:
Instead of trying to read an entire debuggee process region at
once, read up to 4096 pages at a time. This should reduce memory
issues when running against processes that allocate huge amounts of
memory such as modern games.

In addition, slightly optimize the performance if byte alignment is on.

### DIFF
--- a/plugins/BinarySearcher/DialogBinaryString.cpp
+++ b/plugins/BinarySearcher/DialogBinaryString.cpp
@@ -108,7 +108,7 @@ void DialogBinaryString::doFind() {
 						}
 
 						// update progress bar every 64KB
-						if ((uint64_t(p) & 0xFFFF) == 0) {
+						if ((reinterpret_cast<uint64_t>(p) & 0xFFFF) == 0) {
 							ui.progressBar->setValue(util::percentage(i, regions.size(), p - &pages[0], region_size));
 						}
 
--- a/plugins/BinarySearcher/DialogBinaryString.cpp
+++ b/plugins/BinarySearcher/DialogBinaryString.cpp
@@ -68,63 +68,72 @@ void DialogBinaryString::doFind() {
 	auto results = new DialogResults(this);
 
 	const edb::address_t align = ui.chkAlignment->isChecked() ? 1 << (ui.cmbAlignment->currentIndex() + 1) : 1;
-	const int sz = b.size();
-	if (sz != 0) {
-		edb::v1::memory_regions().sync();
-		const QList<std::shared_ptr<IRegion>> regions = edb::v1::memory_regions().regions();
-		const size_t page_size                        = edb::v1::debugger_core->pageSize();
+	const size_t sz = b.size();
+	edb::v1::memory_regions().sync();
+	const QList<std::shared_ptr<IRegion>> regions = edb::v1::memory_regions().regions();
+	const size_t page_size = edb::v1::debugger_core->pageSize();
 
-		int i = 0;
-		for (const std::shared_ptr<IRegion> &region : regions) {
-			const size_t region_size = region->size();
+	int i = 0;
 
-			// a short circut for speading things up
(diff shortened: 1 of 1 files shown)


plugins/BinarySearcher/DialogBinaryString.cpp
```diff
@@ -84,30 +85,35 @@ void DialogBinaryString::doFind() {
...
-					// compare values..
-					if (std::memcmp(p, b.constData(), sz) == 0) {
-						const edb::address_t addr  = p - &pages[0] + region->start();
-						const edb::address_t align = 1 << (ui.cmbAlignment->currentIndex() + 1);
-
-						if (!ui.chkAlignment->isChecked() || (addr % align) == 0) {
+			// Read out 4096 pages at a time, as some applications have huge regions
+			// and we will push against memory fragmentation if we mirror those regions.
+			// To prevent missing a needle in the haystack if it is split between read
+			// boundaries. increment current_page by only 4095.
```
10110111, line 91: I guess you mean a comma instead of a period.

More to the point. What if the needle is longer than 4K? I think this algorithm will still miss it.

plugins/BinarySearcher/DialogBinaryString.cpp
```diff
@@ -84,30 +85,35 @@ void DialogBinaryString::doFind() {
...
 							results->addResult(DialogResults::RegionType::Data, addr);
 						}
-					}
 
-					// update progress bar every 64KB
-					if ((uint64_t(p) & 0xFFFF) == 0) {
-						ui.progressBar->setValue(util::percentage(i, regions.size(), p - &pages[0], region_size));
-					}
+						// update progress bar every 64KB
+						if ((uint64_t(p) & 0xFFFF) == 0) {
```
10110111, line 111: This should be an explicit `reinterpret_cast`.

plugins/BinarySearcher/DialogBinaryString.cpp
eteran: Perhaps we can have it pick a size that is the size of the needle rounded up to nearest multiple of 4096?

plugins/BinarySearcher/DialogBinaryString.cpp
eteran: Agreed, we prefer c++ style casts as they stand out more in the code.

eteran:
Thank you for the PR! 

I generally agree with @10110111's comments, but do love the idea of making things more memory efficient where they can be.

If you can address the concerns he mentioned, I think we'll likely be happy to merge in your contribution 👍

plugins/BinarySearcher/DialogBinaryString.cpp
AaronOpfer: Do we actually expect anyone to use the binary search feature on a string larger than 4K pages? At a minimum this is 4096 bytes if the page size was one, but realistically we're looking at 4KiB pages meaning we're not bugged unless we exceed 16MiB.

10110111: > Do we actually expect anyone to use the binary search feature on a string larger than 4K pages? At a minimum this is 4096 bytes if the page size was one, but realistically we're looking at 4KiB pages meaning we're not bugged unless we exceed 16MiB.

Actually, it was my mistake: I though about 4K bytes rather than pages. Indeed, it should have been 4K pages, which wouldn't make sense for the entry fields in the dialog. Maybe all that needed to be done was setting a limit on the length of these fields (`maxLength`), just in case someone pastes such a large piece of data.

eteran: @10110111, so do you think it's ready to merge then?

10110111: Yep.

PR #824, opened 2022-09-13
Fix build with Capstone 5

thesamesam:
Capstone 5 drops X86_INS_VCVTPD2DQX and X86_INS_VCVTPD2PSX from x86.h (happened in https://github.com/capstone-engine/capstone/commit/dd91f03fef5102492352756e2adb5972dd79a70f but no real explaantion).

Given:
1. these instructions had no wiring up;
2. other Capstone consumers have done the same let's drop based on a macro for the Capstone version.

Bug: https://bugs.gentoo.org/843755
Signed-off-by: Sam James <sam@gentoo.org>


10110111:
Given that these instruction names already puzzled me when I first added them to this `switch` (Intel has no docs for them, nor does a web search reveal any), I wonder whether they have actually ever been emitted by Capstone. Maybe these lines should be simply removed?

thesamesam:
That sounds like a good idea. I was being conservative just in case the FIXME comment was obsolete, but I don't see any benefit to keeping these here at all. Let me update this to drop them.

eteran:
LGTM. Thanks for the patch!

thesamesam:
Thanks!

PR #843, opened 2023-10-12
Support for XFCE terminal

D0m0:
Parameters for xfce4-terminal are a bit different


eteran:
Looks good to me! Thanks!

PR #844, opened 2023-10-26
Fix CMP0115 warning and use default DEFAULT_PLUGIN_DIR or accept

tbhaxor:
- In the new version of CMP0115 policy extension must also be provided, there is no version.h in PROJECT_SOURCE_DIRECTORY.
- Accept the DEFAULT_PLUGIN_DIR during CMAKE configure and use default value to ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/edb, if not provided.


eteran:
LGTM

PR #859, opened 2024-03-24
Show which bookmarks have breakpoints

Talndir:
New features:
* Bookmarks now show their symbol name alongside their address, if it can be found.
* Bookmarks have the same icons beside them that addresses do in the disassembly view (breakpoint, current instruction, or both).
* While bookmark view is in focus, "B" will toggle a breakpoint at that address, and "Shift+B" will add a conditional breakpoint.

It's not all done in the optimal way, but I didn't want to change too many things outside of the Bookmark plugin (the only other change I made is a function in the `edb::v1` namespace that returns the current instruction pointer address).


eteran:
LGTM, thanks for your contribution!

PR #869, opened 2024-07-24
Run lupdate on plugins too

10110111:
This makes plugins translatable. The full command is:
```
lupdate src plugins -ts ./src/res/translations/edb_zh_CN.ts
```

### DIFF
--- a/src/res/translations/edb_zh_CN.ts
+++ b/src/res/translations/edb_zh_CN.ts
@@ -2,66 +2,1182 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="zh_CN" sourcelanguage="en_US">
 <context>
+    <name>AnalyzerPlugin::Analyzer</name>
+    <message>
+        <location filename="../../../plugins/Analyzer/Analyzer.cpp" line="190"/>
+        <source>Analyzer</source>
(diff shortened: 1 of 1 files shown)


eteran:
This definitely will work correctly for first party plugins. One thing we will probably need to consider is how 3rd party plugins interact with translations (if at all). Do we eventually allow translation files to be embedded in the plugin themselves so they can be loaded as well?

10110111:
> Do we eventually allow translation files to be embedded in the plugin themselves so they can be loaded as well?

One way may be to query all the plugins for their own `QTranslator`s and install them after (or before?) the app's ones. Not sure how this would work for conflicting strings translated differently.

eteran:
Yeah, something like this a good idea. We'll have to experiment with it.

eteran:
@10110111 are you thinking that this one should be merged too?

10110111:
I think it's a good start to get something at least all the stock GUI translatable. If you come up with a way to make external plugins translatable, this can be done in subsequent commits, with the strings being already present (assuming you get your translators to fill them in), just waiting for possible splitting.

eteran:
cool, i'll merge this in then :-)

Thanks as always!

PR #870, opened 2024-07-24
Make register categories not translatable

### LINKED ISSUE #868: Unable to display the value of registers such as rax
When debugging a program on Ubuntu on x86-64, the values ​​of registers such as rax and rbx cannot be displayed. Is there any way to display them? Thanks!
image
I can't right-click to display registers such as rax.
image

10110111:
They are not user-visible, only used as ids for searching, so translation is an unnecessary complication that can break things.

Fixes #868


10110111:
Actually wait, this isn't sufficient. There's some problem with rFLAGS, it seems to be `tr`ed somewhere else.

eteran:
No worries, I won't merge until you say so 👍🏻

10110111:
OK, now it seems to work.

eteran (approved):

PR #885, opened 2025-12-09
hardware breakpoint dialog

alex-botw:
I need this code to disable just one hardware breakpoint in the dialog. Now I can only disable all of them at once.


alex-botw:
Do we need to clear DR7 from breakpoint types?

Issue #810, opened 2021-11-06
Labels not saved in session files


cgimenez:
New edb user here, so forgive me if I missed something.

This question is stated here https://github.com/eteran/edb-debugger/issues/98
"Eventually, labels should be saved and loaded with sessions, but not yet."

Losing labels is really a problem, but is it a bug or a not implemented functionality ?

I had a look to the session file and can see that only bookmarks and comments are saved.

eteran:
I believe that they are not yet saved, but you are correct that they should be. We'll use this issue to track it's progress :-)

eteran:
Implemented! (and ASLR aware)

Issue #762, opened 2020-05-03
Persistent breakpoints in base executable


ttufts:
It would be amazing for breakpoints to persist across runs of edb for the base executable.

Vic063:
Hi,

I just created PR #763 which addresses your request as I need the same behavior of EDB.

Regards.

sayboltm:
Sessions, or the import/export breakpoints to work, but sessions would be a true enhancement. Opening an issue now!

eteran:
Implemented! (and ASLR aware)

Issue #358, opened 2016-04-02
Preserve breakpoints across debuggee restart


10110111:
Pressing Ctrl+F2 (or choosing Debug->Restart) currently is (mostly) equivalent to killing the process and launching it again. In particular, the breakpoints are lost on such restart. It'd be useful if they could be preserved.

eteran:
Right. Now that we have session files this is doable, but has a little bit of trickiness to it:

1. ASLR, we need to make sure to store breakpoints as offsets from a region's start instead of absolute addresses. We probably should store a hash of each region that has a BP so we can detect if it changed and therefore should not restore BPs.

2. This is the trickier part. Not all libraries are loaded on start, and with the ptrace API, there is no "event" for library was just loaded like there is on Windows... But there is a solution.

By the time we get to "main", there is a "debug pointer" that is set by ld (see `Debugger::next_debug_event` for some code which attempts to identify it).

One of the things that this pointer refers to is a place in memory which will be jumped to via a `call` instruction on each library load event. So we need to place an internal BP there and treat hitting that as a library load event. It's not pretty, but it'll work :-).

eteran:
So once we have the library load hooked, my current thoughts for how to store breakpoints would look like this:

    {
        [
            "module" : "/lib64/ld-2.21.so",
            "checksum" : "<MD5>",
            "breakpoint_offsets" : [ "1234", "5678", "9abc" ... ]
        ]
    }

Addresses in hex as strings so we don't lose precision converting between floats and values. This would support multiple regions and their offsets and we can detect if the module changed (letting us warn during restoration).

I think this'll cover our needs, but first thing's first, gotta finish implementing load library hooks on linux.

10110111:
What do you mean by floats? Addresses are integers.

eteran:
Exactly. But JSON numbers according to the JSON standard are floating point. So to prevent any unexpected loss of precision, we should store all addresses as strings and convert back to integer values when we read it back.

ojura:
+1 for adding this.

Anutrix:
Any updates on this?

redthing1:
Any update on this?

eteran:
It's been a while for sure, I would love to get this done, but life has been particularly busy. We'll get there eventually though :-).

eteran:
@CherryDT (and others), I certainly want to get this in there. I've even been considering an improved session API which would make use of sqlite DB instead of just storing a JSON blob.

[...]

PR #906, opened 2026-07-14
Bookmark save/restore with library load awareness!

eteran:
this is a bit of a brute force approach but works! now bookmarks are saved/restored in all modules

(heap/stack is kinda a wild card, but not too much we can do about that).

Only real annoyance is minor, and that is that they aren't restored upon _start, but only once ld has loaded the modules they reside in. Might be able to hack around that though...

### DIFF
--- a/include/Comment.h
+++ b/include/Comment.h
@@ -10,8 +10,8 @@
 #include <optional>
 
 struct Comment {
-	edb::address_t address;
 	QString comment;
+	edb::address_t address;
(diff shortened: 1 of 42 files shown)


eteran:
Closes All of these!

https://github.com/eteran/edb-debugger/issues/527
https://github.com/eteran/edb-debugger/issues/810
https://github.com/eteran/edb-debugger/issues/358
https://github.com/eteran/edb-debugger/issues/762

